PR #2, opened 2016-05-09
Minor grammar changes, and a link to GitHub

konklone:
I know you're not asking for grammar fixes on the SP, but I found a few things on the homepage intro, which seemed worth the investment. I won't be offended if you reject it.


pgrassi-nist:
@konklone thanks!

PR #6, opened 2016-05-11
Remove incomplete DOI links

jimfenton:


jimfenton:
I see that you handled this differently in the nist-pages branch, with comment tags. Would you like me to do it that way in master instead?

pgrassi-nist:
Your approach is just fine.  Thanks!

PR #10, opened 2016-05-13
Display documents in a single unit; header fixups

jimfenton:


paul-grassi:
@jimfenton please verify the merge. My phone is wonky.

jimfenton:
It looks good, except the issue I mentioned separately that, by adding the left navigation bar on the home page, I seem to have messed up the right margin alignment somehow (it runs off the right side of the window rather than wrapping properly). I can play around with that a bit more, but we might need the help of someone that knows the NIST templates and CSS.

pgrassi-nist:
I would prefer we go back to cover rather than get nist involved. Or put doc icons on top.

-------- Original Message --------

PR #17, opened 2016-05-16
Rendering edits

jimfenton:
Reference timestamps don't seem to be working yet, but aren't causing any problems.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+**Organization**:
```
paul-grassi, line 1: @jimfenton why is this being ported?  I thought it has to live in master?  Or does it just have to live in default branch?

jimfenton, line 1: [GitHub documentation](https://help.github.com/articles/creating-an-issue-template-for-your-repository/) says this needs to be in the default branch, so this is in preparation for changing the default to nist-pages.

PR #75, opened 2016-06-06
Spelling corrections, s/applicants/applicant’s/

saltybeagle:


paul-grassi:
@saltybeagle Thanks for the edits.  We are using nist-pages as the master, so I will apply these changes there.  We need to delete master...

PR #101, opened 2016-06-21
Updated DC requirements

paul-grassi:
@hferraio, what do you think?


sp800-63a/sec6_dc.md
```diff
@@ -8,17 +8,17 @@ Where the applicant already possesses recognized authentication credentials from
 
 - Before issuing any derived credential the CSP SHALL verify the original credential status and SHALL verify that the corresponding credential is possessed and controlled by the claimant.  
 - The CSP SHALL record the details of the original credential used as the basis for derived credential issuance. 
-- The CSP SHALL set the expiration of the new credential to the expiration of the original credential.
-- The CSP that issued the derived credential SHOULD notify the issuer of the original credential if the derived credential is revoked.
+- The CSP SHOULD set the expiration of the new credential to the expiration of the original credential. There are instances, such as loss of primary credential, where the derived credential need not be directly tied to the expiration of the primary credential.
```
hferraio: The CSP SHOULD set the expiration of the new credential to the expiration of the original credential. There are instances, where the derived credential need not be directly tied to the expiration of the primary credential as the derived credential can provide authentication services in  its place  -- while the expiring primary credential is being replaced.
 12   +- The CSP that issued the derived credential SHOULD notify the issuer of the original, primary credential if the derived credential is revoked.

PR #231, opened 2016-08-12
Appendix addition for password complexity and adding / clarifying examples for out of band

SyntaxPolice:
A few recommendations in case they are helpful:
- Added a summary to the appendix to clarify argument for complexity cross-referenced in section 5.
- Small typo in appendix (try guess instead of just 'try' or 'guess').
- A minor recommendation to add 'chat' as another example (besides VOIP) due to the recent prevalence of chat services that integrate with / synchronize with SMS (e.g. Google Hangouts/iMessage).


sp800-63b/sec5_authenticators.md
```diff
@@ -112,7 +112,7 @@ If the out of band authenticator sends an approval message over the secondary co
...
+Due to the risk that SMS messages may be intercepted or redirected, implementers of new systems SHOULD carefully consider alternative authenticators. If the out of band verification is to be made using a SMS message on a public mobile telephone network, the verifier SHALL verify that the pre-registered telephone number being used is actually associated with a mobile network and not with a VoIP, chat, or other software-based service. It then sends the SMS message to the pre-registered telephone number. Changing the pre-registered telephone number SHALL NOT be possible without two-factor authentication at the time of the change.  **OOB using SMS is deprecated**, and may no longer be allowed in future releases of this guidance.
```
jimfenton, line 115: I'm not clear on what you mean by "chat". If you're referring to services like iMessage that use telephone numbers for addresses, there isn't an association to a telephone number that can be looked up as one can to see if it is a mobile number. The sender just needs to send via SMS and not the chat service (although one could imagine a chat service that is bound to a specific device, and might therefore be acceptable).

SyntaxPolice, line 115: I'm referring to offerings by [Apple](https://support.apple.com/kb/PH22363?locale=en_US) and [Google](https://support.google.com/voice/answer/6023920) for example that integrate chat services with SMS services so that the OTPs are relayed to e.g. the user's browser in addition to being sent via SMS.

Services like Google Voice clearly fall under the "VoIP" service that are already covered, but chat services like iMessage aren't VoIP, but nevertheless seem to have exactly the same issue that the "VoIP" reference seems to be attempting to address. Similarly, systems like [Project Fi](https://support.google.com/fi/answer/6159953?hl=en) are virtual mobile operators, and deliver your text messages to Hangouts, which is a chat service that works across any logged-in device.

jimfenton, line 115: The problem is that this introduces an impossible requirement: there is no way for the verifier to check if the phone number is forwarded to something other than the telephone network. But this is indeed another weakness in the use of SMS as proving possession and control of a specific authenticator.

At the same time, there are chat services, like perhaps Signal, that permit only one endpoint and detect and warn if that endpoint has changed. I can't say for sure that meets all of the requirements, but it's close enough that the specification should not rule out chat services entirely.

SyntaxPolice, line 115: I think you make excellent points about how the verifier _can_ perform validation of an SMS-based out of band factor. Perhaps MDMs could control this by restricting configuration options for forwarding SMS and installation of apps that read SMS, etc.

I didn't intend for my pull request to introduce a new requirement. My interpretation is that it's already present in the text, which currently reads "the verifier SHALL verify that the pre-registered telephone number being used is actually associated with a mobile network and not with a VoIP (_or other software-based_) service" [emphasis added]

If verifiers must check whether the number is associated with VoIP or another software service, _chat_ illustrates an example of software that's extremely common today (see examples above), albeit often through a different mechanism (relay).

With regard to your point about signal, I'm not clear that its chat service would fall under the SMS guidance. It seems to me that this would be a different type of out of band secret?

SyntaxPolice:
After discussion, added #244 to replace this PR.

PR #244, opened 2016-08-19
Added cleaned up summary to tie back to section 5 cross-reference.

SyntaxPolice:
This is a cleaned up PR to address whitespace problem in #231 and remove the unwanted commit related to SMS authenticators. @jimfenton please use this one instead.


jimfenton:
Thanks; this is much easier than modifying #231 as we had discussed.

PR #283, opened 2016-08-22
Typo fix

cybersins:
Removed 'The CSP' typed twice.


BuckyGoat:
> To manage this phase properly, we sincerely request that commenters provide substantive input. Editorial or general comments will be accepted begrudgingly. We want substantive technical/procedural comments. We’ll get to the grammar and formatting later. Trust us, this is as hard for some of the rather strict grammarians here at NIST as it is for anyone!

[NIST Website](https://pages.nist.gov/800-63-3/)

cybersins:
Thanks for the excerpt. I will keep that in mind.

paul-grassi:
We will always accept pull requests, so you can re-submit if you like.

PR #320, opened 2016-08-31
B issues 5

jimfenton:


sp800-63b/sec6_lifecycle.md
```diff
@@ -6,7 +6,7 @@ During the lifecycle of an authenticator bound to a subscriber's identity, a num
 
 ### <a name="binding"></a>6.1. Authenticator binding
 
-Authenticators may be provided by a CSP as part of a process such as enrollment; in other cases, the subscriber may provide their own, such as software or hardware cryptographic modules. For this reason, we refer to the *binding* of an authenticator rather than the issuance, but this does not exclude the possibility that an authenticator is issued as well.
+Authenticators may be issued (provided) by a CSP as part of a process such as enrollment; in other cases, the subscriber may provide their own, such as software or hardware cryptographic modules. For this reason, this guideline refers to the *binding* of an authenticator rather than the issuance, but this does not exclude the possibility that an authenticator is issued as well.
```
paul-grassi: Capitalize MAY?

sp800-63b/sec6_lifecycle.md
```diff
@@ -33,30 +33,21 @@ Enrollment and binding MAY be broken up into a number of separate physical encou
 
 #### 6.1.2. Post-Enrollment Binding
 
-Following enrollment, binding an additional authenticator to an account requires the use of an existing authenticator of the same type (or types). For example, binding a new single-factor OTP device requires the subscriber to authenticate with another *something you have* authentication factor. If the account has only one authentication factor bound to it (which is possible only at IAL 1/AAL 1), an additional authenticator of the same factor may be bound to it.
+Following enrollment, CSPs SHOULD permit the binding of additional authenticators to a subscriber's account. In order to do so, the CSP SHALL require the subscriber to first authenticate at the AAL of the authenticator to be added. When an authenticator is added, the CSP SHOULD send a notification to the subscriber.
```
paul-grassi, line 36: I am not sure we can state this, but I will reserve for our discussion, because the following requirements seem to achieve what we are trying to do.  Just not sure yet...

sp800-63b/sec6_lifecycle.md
```diff
@@ -33,30 +33,21 @@ Enrollment and binding MAY be broken up into a number of separate physical encou
...
 
-Binding an additional authenticator SHALL require the use of two different authentication factors, except as provided below.
-
-If the subscriber has only one of the two authentication factors, they SHALL repeat the identity proofing process, using the remaining authentication and SHOULD verify knowledge of some information collected during the proofing process to bind to the existing identity. In order to reestablish authentication factors at IAL 3, they SHALL verify the biometric collected during the proofing process.
-
-***Consider what proofing information the CSP is allowed to maintain. Privacy impact here?***
+If the account has only one authentication factor bound to it (which is possible only at IAL 1/AAL 1), an additional authenticator may be added, and if it is a different factor (e.g., *something you have* added when authenticated with *something you know*), the subscriber MAY request that the account be upgraded to AAL 2 (but still at IAL 1). Once this has been done, it SHALL no longer be possible for the subscriber to use single-factor authentication.
```
paul-grassi: we shift from 'authentication factor' to authenticator.  I think this will get confusing.  Perhaps, 'a single-factor authenticator'? And request seems iffy.  'the subscriber MAY be allowed to add the factor, therefore upgrading the account.'

sp800-63b/sec6_lifecycle.md
```diff
@@ -33,30 +33,21 @@ Enrollment and binding MAY be broken up into a number of separate physical encou
...
 
+If the subscriber loses one of the two authentication factors and has been identity proofed at IAL 2 or 3, they SHALL repeat the identity proofing process, using the remaining authenticator and SHOULD verify knowledge of some information collected during the proofing process to bind to the existing identity. Reestablishment of authentication factors at IAL 3 SHALL be done in person and SHALL verify the biometric collected during the proofing process.
```
paul-grassi, line 40: Repeat the process, or just prove possession and perform KBA?  That's how it reads to me.

sp800-63b/sec6_lifecycle.md
```diff
@@ -33,30 +33,21 @@ Enrollment and binding MAY be broken up into a number of separate physical encou
...
-
-{:/comment}
+In some instances, a claimant may already possess authenticators at a suitable AAL without having been proofed at the equivalent IAL. For example, a user may have a two-factor authenticator from a social network provider, considered AAL2 and IAL1, and would like to use those credentials at a relying party that requires IAL2.
 
 The following requirements apply when a claimant choses to increase IAL in order to bind to a suitable authenticator they already have.
 
-1. The CSP MAY accept an existing authenticator at or above the desired IAL
-2. The CSP SHALL require the user to authenticate using their existing authenticator
+1. The CSP MAY accept existing one or more existing authenticators
+2. The CSP SHALL require the user to authenticate using their existing authenticator(s) at an AAL compatible with the desired IAL (as defined in [SP 800-63-3](sp800-63-3.html), Acceptable IAL and AAL Combinations table)
```
paul-grassi, line 49: Much better :) I'm wondering now where we talk about assertions needing to carry IAL and AAL values...?

jimfenton: They should probably carry IAL, but AAL is inherited from the authentication itself.

paul-grassi: How is that conveyed in the assertion?  In saml its AuthNContext and acr in OIDC, but we should consider at least a stating that they both need to be sent so an RP can make a decision...

jimfenton: Even if there are no other assertions being conveyed to the RP, the result of authentication is the assertion of an identifier that allows the RP to associate this session with previous authenticated sessions from the same user. That identifier needs an assurance level too, which is the AAL.

sp800-63b/sec6_lifecycle.md
jimfenton: The choice of wording is intentional. Although not a common case, an account could have multiple authenticators of the same factor: multiple physical authenticators or perhaps even multiple memorized secrets. So multiple authenticators doesn't necessarily mean multiple authentication factors. But the wording could be clearer, perhaps _If the account has authenticators of only one authentication factor bound to it_ . Does that help?

PR #328, opened 2016-09-02
Minor edits and issue closer.

### LINKED ISSUE #305: Address Confirmation
**Organization**: TransUnion

**Type**: 2

**Document (63-3, 63A, 63B, or 63C)**: 63A

**Reference (Include section and paragraph number)**: 4.5.6; bullet 7

**Comment (Include rationale for comment)**:
Current guidance "A notification of proofing SHALL be sent to **different** address of record than the destination of the enrollment code. The address of record MAY include the postal address obtained from validated identity evidence. For example, if the CSP sends an enrollment code to a mobile phone of record, a notification of proofing will be sent to the postal address in records or obtained from validated evidence, like a drivers license."

If a notification is required, and it is to go to an address that is different than previously used for an enrollment code, it would seem that sometimes this rule can not be followed. For example, if an enrollment code is forwarded to a consumer's home address to complete the proofing process, and they do not have a different address to then send a notification of proofing, then the guidance would appear to be impossible to follow.  Perhaps I am missing something.

**Suggested Change**:
Remove the SHALL requirement for notification. This is an expense and industry can deliver assurance levels high enough that it adds nothing to security. If it must stay, then modify guidance to allow for eventualities that include the real possibility that a consumer may not have a cell phone or may not choose to allow an enrollment code to be transmitted via SMS.

---

Organization: 1 = Federal, 2 = Industry, 3 = Other

paul-grassi:
Closes #305


sp800-63a/sec5_proofing.md
```diff
@@ -47,7 +47,7 @@ The following table lists qualities, ranging from unacceptable to superior, of i
 
 |Strength|Method(s)|
 |:---:|:------------------------------| 
-|Unacceptable|Validation of the evidence failed.|
+|Unacceptable|No validation or validation of the evidence failed.|
```
jimfenton: Reads awkwardly. Perhaps either add comma after first _validation_ or change to "Validation of the evidence was not performed or failed."

Analogous comment on next diff chunk.

sp800-63a/sec5_proofing.md
```diff
@@ -62,7 +62,7 @@ Once identity evidence is obtained by the CSP, the accuracy, authenticity, and i
 
 |Strength|Method(s)|
 |:---:|:------------------------------| 
-|Unacceptable|No validation or validation of the evidence failed.|
+|Unacceptable|Evidence validation was not performed, or validation of the evidence failed.|
```
paul-grassi, line 65: @jimfenton fixed here and below

PR #329, opened 2016-09-02
Clarify requirements on post-enrollment binding

jimfenton:


paul-grassi:
Looks awesome.  Let's have Hildy look at the IAL3 recovery language.  I think FIPS 201 is a little softer (or maybe uses different terms).  Want to make sure we are good there.

PR #344, opened 2016-09-07
Assertion federation privacy updates

jricher:
Combines and rebases several pull requests. 

closes #336
closes #337 
closes #338 
closes #339 
closes #340


sp800-63c/sec6_presentation.md
```diff
@@ -2,7 +2,7 @@
 
 ## 6. Assertion Presentation
 
-Assertions MAY be presented in either an *indirect* or *direct* manner from the IdP to the RP. Each model has its benefits and drawbacks, but both require the proper validation of the assertion. Assertions MAY also be proxied to facilitate federation between IdPs and RPs under specific circumstances, as discussed in section 4.1.4.
+Assertions MAY be presented in either an *back-channel* or *front-channel* manner from the IdP to the RP. Each model has its benefits and drawbacks, but both require the proper validation of the assertion. Assertions MAY also be proxied to facilitate federation between IdPs and RPs under specific circumstances, as discussed in section 4.1.4.
```
jimfenton: s/in either an/in either a/

sp800-63c/sec6_presentation.md
```diff
@@ -18,16 +18,25 @@ In the *back-channel* model, the subscriber is given an assertion reference to p
 
 In this model, the assertion itself is requested directly from the IdP to the RP, minimizing chances of interception and manipulation by a third party (including the subscriber themselves). 
 
-This method also allows the RP to query the CSP for additional attributes about the subscriber not included in the assertion itself, as back-channel communication can occur after the initial authentication transaction. 
+This method also allows the RP to query the CSP for additional attributes about the subscriber not included in the assertion itself, as back-channel communication can occur after the initial authentication transaction, or when the subscriber is no longer available.
```
jimfenton: "when the subscriber is no longer available". I'm wondering whether this is in-scope for 800-63, or if at that point we're talking about a machine-to-machine, and not a user-to-machine transaction.

sp800-63c/sec5_strength.md
```diff
@@ -70,12 +74,14 @@ All assertions SHOULD use audience restriction techniques to allow an RP to reco
 
 #### 5.2.5. Pairwise Pseudonymous Identifiers
 
-In some circumstances, it is desirable to prevent the subscriber's account at the IdP from being linked through one or more RPs through use of a common identifier. Pairwise pseudonymous subject identifiers SHALL be used within the assertions generated by the IdP for the RP, and the IdP SHALL generate a different identifier for each RP or integration with an RP, as appropriate.
+In some circumstances, it is desirable to prevent the subscriber's account at the IdP from being linked through one or more RPs through use of a common identifier. In these circumstances, pairwise pseudonymous subject identifiers SHALL be used within the assertions generated by the IdP for the RP, and the IdP SHALL generate a different identifier for each RP.
```
jimfenton, line 77: Do we need to say something about how the decision to use pairwise pseudonymous identifiers is made? Is it an RP decision, a user decision, or both?

sp800-63c/sec6_presentation.md
```diff
@@ -18,16 +18,25 @@ In the *back-channel* model, the subscriber is given an assertion reference to p
...
+ - SHOULD be time limited with a short lifetime of seconds or minutes
+ - SHOULD be presented along with authentication of the RP
 
 The RP SHALL protect itself against injection of manufactured or captured assertion references by use of cross-site scripting protection or other accepted techniques. 
 
 Claims within the assertion SHALL be validated including issuer verification, signature validation, and audience restriction.
 
 Conveyance of the assertion reference from the IdP to the subscriber as well as from the subscriber to the RP SHALL be made over an authenticated protected channel. Conveyance of the assertion reference from the RP to the IdP as well as the assertion from the IdP to the RP SHALL be made over an authenticated protected channel.
 
+The assertion reference SHOULD require authentication of the client upon presentation at the IdP before an assertion is issued.
```
jimfenton: What is the 'client' here? This is the first use of the term. Also, (although not really part of this commit), I'm confused by the discussion involving the term IdP when there is no IdP shown in Figure 1.

sp800-63c/sec6_presentation.md
```diff
@@ -38,11 +47,11 @@ In the *front-channel* model, the IdP creates an assertion and sends it to the s
 
 In the front-channel method, an assertion is visible to the subscriber, which could potentially cause leakage of system information included in the assertion. 
 
-Since the assertion is under the control of the subscriber, the direct method also allows the assertion to be replayed the subscriber to submit a single assertion to multiple RPs. This
+Since the assertion is under the control of the subscriber, the direct method also allows the subscriber to submit a single assertion to unintended parties, perhaps by a browser replaying an assertion at multiple RPs. Intentional use of this method can lead to lax audience restriction of the assertion itself, which in turn could lead to privacy and security breaches for the subscriber across these RPs. Such multi-RP use is not recommended. Instead, RPs are encouraged to fetch their own individual assertions.
```
jimfenton: Still using the term _direct method_. Also confused how an audience-restricted assertion would be useful at unintended parties.

sp800-63c/sec6_presentation.md
jricher: This is definitely in scope as it's part of how the federation protocol can work. It is a machine to machine transaction but one about the current user, and I think it's important that we note it as part of the authentication process here. I also don't think we need to profile it beyond mentioning it like this.

sp800-63c/sec6_presentation.md
jricher: "client" should be "RP".

Diagrams weren't updated when we switched to using IdP in this document, they need to be redone and re-rendered.

sp800-63c/sec6_presentation.md
jricher: If it's restricted to multiple audiences (looseness on the server side) or the RP doesn't do the right checks (looseness on the client side) then this happens. It's not supposed to, of course, but in practice we see this a lot as RP deployers try to cut corners and get one assertion to use everywhere.

sp800-63c/sec5_strength.md
jricher: Possibly, but that decision varies depending on the protocol so I left that text out.

PR #346, opened 2016-09-08
Change "htmlonly" to "httponly"

jricher:
Perhaps also add a reference to https://www.owasp.org/index.php/HttpOnly somewhere?


jimfenton:
Good catch; I mix up HTML and HTTP all the time.

With respect to the OWASP link, I'm a little concerned about including a link that may not be stable, and this is a wiki page. Is this a source that would be considered authoritative? We could also reference a standard that defines httponly, but that probably wouldn't say enough about the motivation for doing so.

jricher:
This seems to be the standard of record on that attribute (but I could be wrong):

https://tools.ietf.org/html/rfc6265#section-4.1.2.6

Issue #189, opened 2016-07-29
Man in the middle definitions 


FIDOAlliance:
FIDO Alliance

800-63-3B 
section 5.1.3

This 800-63-3 mentions resistance to man-in-the-middle attacks.  The older 800-63-2 mentioned weak man-in-the-middle resistance and strong man-in-the-middle resistance.

Does the term in 63-3 relate to the old “weak” or “strong” MITM resistance (or is it different)? 

---

Organization:  2 = Industry

jimfenton:
The language around MITM resistance will be clarified. Rather than the term "weak man-in-the-middle resistance" (which seems, well, weak) we will be distinguishing between man-in-the-middle resistance and authentication that binds more strongly to the server's identity, tentatively to be referred to as phishing resistance.

jimfenton:
Current thought is to refer to "strong man-in-the-middle resistance" as "verifier impersonation resistance". Proposed text around this is being reviewed, and can be seen at https://github.com/jimfenton/800-63-3/commit/7ea140ddea1c4747aa460bc4655ab8337abefd9e

jimfenton:
Resolved by #371.

PR #371, opened 2016-09-14
Add verifier impersonation resistance requirements

jimfenton:


jimfenton:
Resolves #189.

sp800-63b/sec3_definitions.md
```diff
@@ -316,8 +316,8 @@ Per ISO/IEC 9241-11: Extent to which a product can be used by specified users to
 #### Verifier
 An entity that verifies the claimant’s identity by verifying the claimant’s possession and control of one or two authenticators using an authentication protocol. To do this, the verifier may also need to validate credentials that link the authenticator(s) and identity and check their status.
 
-#### Verifier Impersonation Attack
-A scenario where the attacker impersonates the verifier in an authentication protocol, usually to capture information that can be used to masquerade as a subscriber to the real verifier.
+#### Verifier Impersonation
+A scenario where the attacker impersonates the verifier in an authentication protocol, usually to capture information that can be used to masquerade as a subscriber to the real verifier. In previous editions of SP 800-63, authentication protocols that are resistant to verifier impersonation have been described as "strongly man-in-the-middle" resistant.
```
jricher: The quotes here seem awkward, would it be better to remove them entirely? Or if we want to quote, we should quote the original terminology directly. What's the exact phrasing from 800-63-2 that we're referencing here?

sp800-63b/sec4_aal.md
```diff
@@ -115,18 +119,26 @@ CSPs shall comply with their respective records retention policies in accordance
...
 * Multi-Factor Cryptographic Device
 * Single-Factor Cryptographic Device used in conjunction with Memorized Secret
 
+> Note: OTP devices do not provide verifier impersonation resistance, which is required at AAL 3, so a cryptographic authenticator or device is also required even though the OTP device is a multi-factor device.
+
+All cryptographic devices and authenticators used at AAL 3 SHALL be resistant to verifier impersonation attacks as described in Section [5.2.5](#verifimpers).
+
 #### 4.3.2. Authenticator and Verifier Requirements
 
+Communication between the claimant and channel SHALL be via an authenticated protected channel to provide confidentiality of the authenticator output and resistance to man-in-the-middle attacks. At least one authenticator used at AAL 3 SHALL be
```
jricher: This sentence trails off. What's the rest?

sp800-63b/sec5_authenticators.md
```diff
@@ -348,3 +348,15 @@ Biometrics are also used in some cases to prevent repudiation of registration an
 Authenticators that are directly connected to or embedded in endpoints MAY convey attestation information such as the provenance or health and integrity of the authenticator (and possibly the endpoint as well) to the verifier as part of the authentication protocol. If this attestation is signed, the verifier SHOULD validate its signature. This information MAY be used as part of a risk-based authentication decision.
 
 When federated authentication is being performed as described in [SP 800-63C](sp800-63c.html), the verifier SHOULD include any such attestation information in the assertion it provides to the relying party.
+
+#### <a name="verifimpers"></a>5.2.5 Verifier impersonation resistance
+
+Verifier impersonation attacks, sometimes referred to as "phishing attacks", refer to attempts by fraudulent verifiers and relying parties to fool an unwary claimant into authenticating to an impostor website. In previous editions of the Guideline, protocols that are resistant to verifier impersonation attacks were also referred to as "strongly man-in-the-middle resistant".
```
jricher: Is "the Guideline" (caps included) the right term? Elsewhere we say "In 800-63-2" or "in previous editions of this document".

sp800-63b/sec5_authenticators.md
```diff
@@ -348,3 +348,15 @@ Biometrics are also used in some cases to prevent repudiation of registration an
...
+
+Authentication protocols that are verifier impersonation resistant SHALL strongly and irreversibly bind the authenticator output to the hostname, domain, or public key of the certificate of the verifier to which it is sent. This makes the authenticator output useless to the attacker since it is not bound to the legitimate verifier.
```
jricher: This statement presumes certificates as the only solution for strong binding. Perhaps relax that to be more general?

jricher (changes requested):

sp800-63b/sec3_definitions.md
jimfenton: The point here was not to quote, but to set off a term that was used in quite a few places in 800-63-2. Italics would probably be more appropriate.

sp800-63b/sec4_aal.md
jimfenton: oops! Watch for the next commit to the branch.

sp800-63b/sec5_authenticators.md
jimfenton: I'm probably being inconsistent in terms. I wasn't sure about "in previous versions of this document" because this is actually the first version of -63B. I thought that calling it "the Guideline" was more inclusive of the document suite, but 800-63-2 is probably easier to understand.

sp800-63b/sec5_authenticators.md
jimfenton: No, it gives a choice: hostname, domain name, or public key. FIDO U2F authenticators are also verifier impersonation resistant, but use the hostname rather than the public key AFAIK.

jricher: While it does give a choice of those three objects, all three are explicitly tied to "the certificate of the verifier" and not just the verifier itself.

sp800-63b/sec3_definitions.md
jricher: In the definitions, I'd suggest either removing the quotes or quoting "strongly man-in-the-middle resistant" as a phrase, as done elsewhere in this pull.

jimfenton: OK, that makes sense.

sp800-63b/sec5_authenticators.md
jimfenton: Yes; what I meant was that it isn't just tied to client-authenticated TLS, which was the only real example in -2. What other examples of verifier authentication can we cite? Maybe something rooted in DNSSEC, but DANE still deals with certificates.

I'm happy to be more general, but need some suggestions on what direction to go. I considered "public key of the verifier" rather than "public key of the certificate of the verifier", but you still need to have some basis for knowing that's the public key of the intended verifier, not the attacker's clone.

sp800-63b/sec5_authenticators.md
jimfenton: There were several phrases used, but "the guideline" or "this guideline" was most common. However, I agree that "previous editions of SP 800-63" is clearer, so have changed all I can find (including definitions in -3 and 63A) to that wording.

sp800-63b/sec5_authenticators.md
```diff
@@ -348,3 +348,15 @@ Biometrics are also used in some cases to prevent repudiation of registration an
...
 
 When federated authentication is being performed as described in [SP 800-63C](sp800-63c.html), the verifier SHOULD include any such attestation information in the assertion it provides to the relying party.
+
+#### <a name="verifimpers"></a>5.2.5 Verifier impersonation resistance
+
+Verifier impersonation attacks, sometimes referred to as "phishing attacks", refer to attempts by fraudulent verifiers and relying parties to fool an unwary claimant into authenticating to an impostor website. In previous editions of SP 800-63, protocols that are resistant to verifier impersonation attacks were also referred to as "strongly man-in-the-middle resistant".
+
+Authentication protocols that are verifier impersonation resistant SHALL strongly and irreversibly bind the authenticator output to the hostname, domain, or public key of the certificate of the verifier to which it is sent. This makes the authenticator output useless to the attacker since it is not bound to the legitimate verifier.
```
jricher: This section still refers to the "certificate of the verifier" as a normative requirement. This language preemptively selects a technological solution to the problem, which I don't believe is intended.

jricher (changes requested):

sp800-63b/sec5_authenticators.md
jimfenton: I wonder if we're parsing the sentence differently.  How about the following:

Authentication protocols that are verifier impersonation resistant SHALL strongly and irreversibly bind the authenticator output to the public key of the certificate presented by the host to which it is sent, or to that host's verified name or domain.

Hopefully that makes it clearer that the hostname or domain can be used independently of certificates. I have also changed some of the language around the word _verifier_ in order to make it clearer that the authenticator output needs to be bound to the actual host it's not authenticating to. If it's bound to the intended verifier instead the impostor could use the authenticator output successfully.

sp800-63b/sec5_authenticators.md
```diff
@@ -353,7 +353,7 @@ When federated authentication is being performed as described in [SP 800-63C](sp
 
 Verifier impersonation attacks, sometimes referred to as "phishing attacks", refer to attempts by fraudulent verifiers and relying parties to fool an unwary claimant into authenticating to an impostor website. In previous editions of SP 800-63, protocols that are resistant to verifier impersonation attacks were also referred to as "strongly man-in-the-middle resistant".
 
-Authentication protocols that are verifier impersonation resistant SHALL strongly and irreversibly bind the authenticator output to the hostname, domain, or public key of the certificate of the verifier to which it is sent. This makes the authenticator output useless to the attacker since it is not bound to the legitimate verifier.
+Authentication protocols that are verifier impersonation resistant SHALL strongly and irreversibly bind the authenticator output to the public key of the certificate presented by the host to which it is sent, or to that host's verified name or domain. This makes the authenticator output useless to the attacker since it is not bound to the legitimate verifier.
```
jricher, line 356: This works, thank you!

jricher (approved):

PR #397, opened 2016-09-22
Verifier impersonation resistance corrections and clarifications

jimfenton:


sp800-63b/sec4_aal.md
```diff
@@ -72,7 +72,7 @@ The CSP shall comply with their respective records retention policies in accorda
...
+AAL 2 provides higher assurance that the same claimant who participated in previous transactions is accessing the protected transaction or data. At least two different authentication factors are required. Various types of authenticators, including multi-factor software cryptographic authenticators, may be used as described below. AAL 2 also permits any of the authentication methods of AAL 3. AAL 2 authentication requires cryptographic mechanisms that protect the primary authenticator against compromise by the protocol threats for all threats at AAL 1. Approved cryptographic techniques are required at AAL 2 and above.
```
pgrassi-nist: This doesn't account for Bill's recommendation, but that may be coming later.  I only mention it since this text got changed.

sp800-63b/sec4_aal.md
```diff
@@ -130,8 +130,9 @@ Authentication Assurance Level 3 requires the use of one of three kinds of hardw
  * Single-Factor Cryptographic Device
 * Multi-Factor Cryptographic Device
 * Single-Factor Cryptographic Device used in conjunction with Memorized Secret
+* Multi-Factor OTP Device used in conjunction with a single-factor cryptographic device
```
paul-grassi, line 133: Maybe I wasn't following the Ray thread but MFA in conjunct with SF crypto looks like 2-haves and 1-know.  What am I missing?

jimfenton: The second "have" is needed because the MF OTP device isn't verifier impersonation resistant, so we need another authenticator that is. Another way to describe it might be "Single-Factor Cryptographic Device used in conjunction with either Memorized Secret or Multi-Factor OTP Device"

I notice that there's a rendering problem with the sub-bullets at line 129 and 130 so will be committing a fix for that.

I was afraid that one of the pull requests would result in a conflict on the other, so if this looks OK I can do the conflict resolution and merge if you want.

PR #403, opened 2016-09-28
clarified PPID generation

jricher:


paul-grassi:
@nbl-nist take a peek.

nbl-nist:
Let me know what you think of the following two sentences as a replacement for your newly edited sentence. I was trying to think from a potential assessor's point of view. Just to be clear, either construction places the burden for determining the validity of the correlation on the IdP. Is that where you want it? 

An IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if those RPs have a demonstrable, pre-existing relationship that could justify an operational need for the correlation, such as a shared security domain or legal ownership. The RPs SHALL conduct a joint privacy risk assessment before requesting that the IdP generate the same identifier.

jricher:
How about instead of a "joint privacy risk assessment" (which sounds like a specific formal process), we have:

> The RPs SHALL consider the privacy risks associated with requesting a common identifier.

nbl-nist:
We’ve called for privacy risk assessments in other requirements. What’s the difference here?

jricher:
@nbl-nist new text on the pull request, please review.

jricher:
Also: yes, it's up to the IdP to determine the validity of the correlation, always, since it's the IdP that is calculating the PPID.

jricher:
The difference is that it's a group of RPs together and that a joint privacy assessment may be unrealistic or unfeasible for a formal and documented process. 

The IdP MAY always reject a requested grouping, on any grounds.

nbl-nist:
I think the consent/request difference between the two sentences is jarring. Practically speaking, are there use cases where the IdP would initiate the generation and seek consent instead of responding to a request from RPs?

jricher:
We need to capture that all RPs are OK with the correlation, not just that one of them requested it and claims to speak for the group, thus having both.

nbl-nist:
You could say “at the joint request.”

nbl-nist:
I’m not really seeing the issue, if they have the kind of operational relationship that we’re expecting. If they’re not that connected, what’s the point of limiting the exception? Anyway, the whole thing seems like hair-splitting since the previous section doesn’t have a blanket requirement to use PPIDs or at least a privacy risk assessment to justify under what circumstances PPIDs wouldn’t be used. I think this latter requirement should be added.

jricher:
Newest checkin re-added the risk assessment text, but kept the other text for appropriate context and motivation.

sp800-63c/sec5_strength.md
```diff
@@ -74,14 +74,19 @@ All assertions SHOULD use audience restriction techniques to allow an RP to reco
...
 
-If several RPs have an explicit and tight correlation, such as a shared security domain or legal ownership, those RPs MAY be combined into a single logical RP for the purposes of generating a pairwise pseudonymous identifier. The IdP SHALL ensure that only intended RPs are in such a group, otherwise a rogue RP could learn of the pairwise pseudonymous identifier by posing as part of an RP group of which it is not a member.
\ No newline at end of file
+An IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if:
+
+* Those RPs have a demonstrable relationship that justifies an operational need for the correlation, such as a shared security domain or shared legal ownership, and
+* All RPs consent to being correlated in such a manner.
+
+The RPs SHALL conduct a privacy risk assessment to consider the privacy risks associated with requesting a common identifier. The IdP SHALL ensure that only intended RPs are correlated, otherwise a rogue RP could learn of the pseudonymous identifier for a correlation by fraudulently posing as part that correlation.
```
jimfenton: s/part that correlation/part of that correlation/

While you're at it, the comma after "correlated" should be a semicolon.

sp800-63c/sec5_strength.md
```diff
@@ -74,14 +74,19 @@ All assertions SHOULD use audience restriction techniques to allow an RP to reco
...
+Pairwise pseudonymous identifiers SHALL be opaque and unguessable, containing no identifying information about the subscriber. Additionally, the identifiers SHALL only be known by and used by one IdP-RP pair. 
 
-If several RPs have an explicit and tight correlation, such as a shared security domain or legal ownership, those RPs MAY be combined into a single logical RP for the purposes of generating a pairwise pseudonymous identifier. The IdP SHALL ensure that only intended RPs are in such a group, otherwise a rogue RP could learn of the pairwise pseudonymous identifier by posing as part of an RP group of which it is not a member.
\ No newline at end of file
+An IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if:
+
+* Those RPs have a demonstrable relationship that justifies an operational need for the correlation, such as a shared security domain or shared legal ownership, and
+* All RPs consent to being correlated in such a manner.
```
jimfenton: All RPs sharing the identifier? Or is the "sharing the identifier" part obvious?

jimfenton (changes requested):
Couple of mostly grammatical nits.

jimfenton (approved):

nbl-nist:
Under the following language, the default is going to be to not use PPIDs because it's unclear under what circumstance it would be desirable and since the benefit accrues to the individual, who advocates for them? The default should be to use them, with exceptions allowed. 

In some circumstances, it is desirable to prevent the subscriber's account at the IdP from being linked through one or more RPs through use of a common identifier. In these circumstances, pairwise pseudonymous subject identifiers SHALL be used within the assertions generated by the IdP for the RP, and the IdP SHALL generate a different identifier for each RP.

sarahcec:
Yeah, that's a valid concern. What if we replace "In some circumstances" with
"Unless there are mitigating factors"?

Sarah Squire
Engage Identity
http://engageidentity.com

jricher:
Or instead of "In some circumstances", we do "In many circumstances"?

nbl-nist:
I like it, but I still don’t know how to interpret “it is desirable.” Is that a SHOULD concept or a SHALL concept?

PR #404, opened 2016-09-28
Risk based language main doc

paul-grassi:


jimfenton:
Normative keywords (SHALL, etc.) are not capitalized, but that seems to be true elsewhere in 800-63-3 as well so will merge the pull and we can clean up capitalizations later.

PR #407, opened 2016-09-29
Defined symmetric and asymmetric signatures

jricher:


jimfenton (commented):
Suggest using the verb "utilizes" rather than requires. Let's see if I can make that change to your branch.

jimfenton:
At some point we should probably alphabetize the definitions as is done in the other volumes.

PR #409, opened 2016-10-03
Biometrics requirements updates

jimfenton:


paul-grassi:
@jimfenton, do you want me to write the 'scary language' in a comment here so you can grab it and add it via a new commit, or just write it after this pull request gets merged?

jimfenton:
Given the line terminator issues, it might be easier for you to put the text in a comment here and I'll commit it.

paul-grassi:
Will do.

sp800-63b/sec5_authenticators.md
```diff
@@ -326,19 +326,21 @@ Biometrics SHALL be used with another authentication factor (something you know
...
+When the biometric sensor and subsequent processing are not part of an integral unit that resists replacement of the sensor, the sensor (or endpoint with which it is an integral unit that resists sensor replacement) SHALL demonstrate that it is a certified or qualified sensor meeting these requirements by authenticating itself to the processing element.
 
-Testing of the biometric system to be deployed SHALL demonstrate at least 90% resistance to presentation attacks for each relevant attack type (aka species), where resistance is defined as the number of thwarted presentation attacks divided by the number of trial presentation attacks. The biometric system SHALL implement presentation attack protection (PAD).
+Testing of the biometric system to be deployed SHOULD demonstrate at least 90% resistance to presentation attacks for each relevant attack type (aka species), where resistance is defined as the number of thwarted presentation attacks divided by the number of trial presentation attacks. The biometric system SHOULD implement presentation attack protection (PAD).
```
paul-grassi: Can you flip these sentences:  The biometric system SHOULD implement presentation attack protection (PAD).  Testing of the biometric system to be deployed SHOULD demonstrate at least 90% resistance to presentation attacks for each relevant attack type (aka species), where resistance is defined as the number of thwarted presentation attacks divided by the number of trial presentation attacks.

sp800-63b/sec5_authenticators.md
```diff
@@ -326,19 +326,21 @@ Biometrics SHALL be used with another authentication factor (something you know
...
 
-The biometric system SHALL allow no more than 10 consecutive failed authentication attempts. Once that limit has been reached, the claimant SHALL be required to use a different authenticator or to activate their authenticator with a different factor such as a memorized secret.
+>Note: Presentation attack detection MAY be required in future editions of this guideline.
```
paul-grassi: I don't think we want normative language here, but I do like the note.  Maybe: Presentation attack detection is being considered as a mandatory requirement in future editions of this guideline.

sp800-63b/sec5_authenticators.md
```diff
@@ -326,19 +326,21 @@ Biometrics SHALL be used with another authentication factor (something you know
...
+
+The biometric system SHALL allow no more than 3 consecutive failed authentication attempts, or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the claimant SHALL be required to use a different authenticator or to activate their authenticator with a different factor such as a memorized secret.
```
paul-grassi, line 335: Why is the claimant required to use another factor? Why not in some implementations just lock them out?  Maybe for usability purposes, like iPhone, the other factor is good to have, but do we need to require it?  Recommend:  the claimant SHOULD be required to use a different authenticator or to activate their authenticator with a different factor such as a memorized secret. @enewt agrees :)

paul-grassi (changes requested):
I will get you the 'scary' language soon.  Just wanted to get review comments back to you as soon as I could.

sp800-63b/sec5_authenticators.md
jimfenton: Makes more sense that way, yes.

sp800-63b/sec5_authenticators.md
jimfenton: In that case, we might want to change the style at the end of paragraph 1 of section 5.1.3.2, which currently reads:

> **OOB using the PSTN (SMS or voice) is deprecated**, and may no longer be allowed in future releases of this guidance.

(which also says "guidance" rather than "guideline"; I thought I caught all of those!

sp800-63b/sec5_authenticators.md
jimfenton: The SHALL here means that either another factor, or another authenticator, is required in order to successfully authenticate. I am concerned that a SHOULD here might be interpreted that there is some exception where another factor might not be required. Or that some different limit might be arbitrarily chosen.

paul-grassi (approved):

PR #410, opened 2016-10-03
Single-factor crypto software authenticators

jimfenton:


paul-grassi:
@jimfenton should we get Ray/Bill/Andy to look at this?  I missed the discussion so would prefer they do it, but don't think this would be too tough for me to handle either.

jimfenton:
Either way; I thought it would be easier to put the text in and have them review, and revise if needed.

sp800-63b/sec5_authenticators.md
```diff
@@ -201,7 +201,28 @@ The verifier SHALL use approved encryption and SHALL utilize an authenticated pr
...
 If the authenticator output or activation secret has less than 64 bits of entropy, the verifier SHALL implement a throttling mechanism that effectively limits the number of failed authentication attempts an attacker can make on the subscriber’s account as described in [Section 5.2.2](#throttle). A biometric activation factor SHALL meet the requirements of [Section 5.2.3](#biometric_use), including limits on number of successive authentication failures.
     
-#### 5.1.6. Single Factor Cryptographic Devices
+#### 5.1.6. Single Factor Cryptographic Software
+
+<div class="text-left" markdown="1">
+<table style="width:100%">
+  <tr>
+    <td><img src="sp800-63b/media/Single-factor-software-crypto.png" alt="authenticator" style="width: 100px;height: 100px"/></td>
+    <td>A single factor software cryptographic authenticator is a cryptographic key stored on disk or some other “soft” media. Authentication is accomplished by proving possession and control of the key. The authenticator output is highly dependent on the specific cryptographic protocol, but it is generally some type of signed message. The SF software cryptographic authenticator is <i>something you have</i>.</td>
```
paul-grassi, line 210: do we want the quoted 'soft' or can we just say 'some other form of software'?  I suppose we could get in trouble because memory is viable.  What kind of credit would we give for SE or TEE?  TPM would be HW I presume.

sp800-63b/sec5_authenticators.md
```diff
@@ -201,7 +201,28 @@ The verifier SHALL use approved encryption and SHALL utilize an authenticated pr
...
+    <td>A single factor software cryptographic authenticator is a cryptographic key stored on disk or some other “soft” media. Authentication is accomplished by proving possession and control of the key. The authenticator output is highly dependent on the specific cryptographic protocol, but it is generally some type of signed message. The SF software cryptographic authenticator is <i>something you have</i>.</td> 
+  </tr>
+  </table>
+  </div>
+
+
+
+##### 5.1.6.1. Single Factor Cryptographic Software Authenticators
+
+Single factor software cryptographic authenticators encapsulate a secret key that is unique to the authenticator. The key SHALL be stored in the most secure storage available on the device (e.g., keychain storage, trusted platform module, or trusted execution environment if available). The key SHALL be strongly protected against unauthorized disclosure by the use of access controls that limit access to the key to only those software components on the device requiring access.
```
paul-grassi, line 219: my own ignorance, 'secret key' is because this can be symmetric or asymmetric?  So u addressed earlier comment about TEE, but TPM is hardware.  I don't think we can say this.  Even the crypto ops on a smart card are done with a software applet on the chip...

sp800-63b/sec5_authenticators.md
```diff
@@ -351,7 +372,7 @@ Biometrics are also used in some cases to prevent repudiation of registration an
 
 #### <a name="attestation"></a>5.2.4 Attestation
 
-Authenticators that are directly connected to or embedded in endpoints MAY convey attestation information such as the provenance or health and integrity of the authenticator (and possibly the endpoint as well) to the verifier as part of the authentication protocol. If this attestation is signed, it SHALL be signed using a difital signature that provides at least the minimum security strength specified in the latest revision of [[SP 800-131A]](#SP800-131A) (currently 112 bits). Attestation information MAY be used as part of a risk-based authentication decision.
+Authenticators that are directly connected to or embedded in endpoints MAY convey attestation information such as the provenance or health and integrity of the authenticator (and possibly the endpoint as well) to the verifier as part of the authentication protocol. If this attestation is signed, it SHALL be signed using a digital signature that provides at least the minimum security strength specified in the latest revision of [[SP 800-131A]](#SP800-131A) (currently 112 bits). Attestation information MAY be used as part of a risk-based authentication decision.
```
paul-grassi: i'd like to add some items to the 'such as'.  For consideration:
- Security features
- Biometric security metrics
- Biometric performance metrics
- Sensor Modality

paul-grassi (approved):

sp800-63b/sec5_authenticators.md
jimfenton: This is cut-and-paste from 800-63-2 (p.45). The intent of "soft" here, I think, is that the key is accessible from a general-purpose processor, as distinct from a "hardware" authenticator that does nothing other than authentication and possibly some other security functions. This wording doesn't prevent the key from being in volatile memory or other things like flash.

Not much of anything is really hardware now, so I suggest we make the distinction based on whether the whole authenticator (everything that runs on its processor, if any) is subject to FIPS 140 requirements: if so, it's "hardware". Not sure where TPM and TEE fit on that spectrum.

sp800-63b/sec5_authenticators.md
jimfenton: Right, the secret key can be either symmetric or asymmetric. See earlier comment about "hardware" authenticators: Everything that runs on a smart card would be subject to FIPS 140 requirements as a unit, so we would consider that "hardware".

sp800-63b/sec5_authenticators.md
jimfenton: Commit https://github.com/usnistgov/800-63-3/pull/410/commits/ef19a74a10eb76a87a062d88e340bf889b7aca16 below attempts to better define the distinction between HW and SW crypto authenticators.

PR #413, opened 2016-10-07
800-63-3 editorial changes

jimfenton:
These are the (hopefully) small and non-controversial changes. I have some other changes that we should discuss before making.


sp800-63-3/cover.md
```diff
@@ -178,16 +178,16 @@ OMB guidance outlines a five-step process by which agencies should meet their di
...
+This document suite provides guidelines for implementing the third step of the above process.  A new approach for digital authentication solutions is required by these guidelines, separating the individual elements of identity assurance into discrete, component parts. For non-federated systems, agencies will select and combine two individual components, referred to as *Identity Assurance Level (IAL)*  and *Authenticator Assurance Level (AAL)*. For federated systems, a 
 third component, *Federation Assurance Level (FAL)*, is required. 
 
 * IAL refers to the robustness of the identity proofing process and the binding between an authenticator and the records pertaining to a specific individual. 
 * AAL refers to the robustness of the authentication process itself.
-* FAL refers to the robustness of the assertion protocol utilized by the federation to communicate authentication and attribute information (if applicable) to a relying party.
+* FAL refers to the robustness of the assertion protocol utilized by the federation to communicate authentication and attribute information (if applicable) to a separate relying party.
```
paul-grassi: separate from what?  I know IdP, but if the inferred separation (as originally written) isn't good form, then we should be clear when adding 'separate'.

sp800-63-3/cover.md
```diff
@@ -208,13 +208,13 @@ A summary of each of the identity, authenticator, and federation assurance level
...
+**Authenticator Assurance Level 1** - AAL 1 provides single factor authentication, giving some assurance that the same claimant who participated in previous transactions is accessing the protected transaction or data. AAL 1 allows a wide range of available authentication technologies to be employed. It also permits the use of any of the authentication methods of higher authenticator assurance levels, therefore allowing CSPs to allow users to use a higher AAL authenticator to be at AAL 1. Successful authentication requires that the claimant prove through a secure authentication protocol that he or she possesses and controls the authenticator.
```
paul-grassi: For AAL1-3, this doesn't look like the new language you added per Bill's comments.

sp800-63-3/cover.md
```diff
@@ -208,13 +208,13 @@ A summary of each of the identity, authenticator, and federation assurance level
...
 
-**Federation Assurance Level 1** - FAL 1 allows for the subscriber to retrieve and present a bearer assertion directly to the RP in the front channel. The assertion must be asymmetrically signed with an appropriate algorithm.
+**Federation Assurance Level 1** - FAL 1 allows for the subscriber to retrieve and present a bearer assertion directly to the relying party (RP) in the front channel. The assertion must be asymmetrically signed with an appropriate algorithm.
```
paul-grassi, line 217: Don't worry about it, but this is the stuff we will fix later.

sp800-63-3/sec4_model.md
```diff
@@ -10,15 +10,15 @@ The digital authentication model used in these guidelines reflects current techn
...
 
-As part of authentication, mechanisms such as device identity or geo-location may be used to identify or prevent possible authentication false positives. While these mechanisms do not directly increase the authenticator assurance level, they can enforce security policies and mitigate risks. In many cases, the authentication process and services will be shared by many applications and agencies. However, it is the individual agency or application acting as the RP that shall make the decision to grant access or process a transaction based on the specific application requirements.
+As part of authentication, mechanisms such as device identity or geo-location may be used to identify or prevent possible authentication false positives. While these mechanisms do not directly increase the authenticator assurance level, they can aid in enforcing security policies and mitigate risks. In many cases, the authentication process and services will be shared by many applications and agencies. However, it is the individual agency or application acting as the RP that shall make the decision to grant access or process a transaction based on the specific application requirements.
```
paul-grassi, line 21: I see 'may'.  Do we need the hassle of cleaning out 'normative' words or is stating these sections (actually all of -3) is informative enough?  I hope the latter :)

paul-grassi (changes requested):

sp800-63-3/cover.md
jimfenton: Maybe, "...to a relying party that is separate from the federation" ?

paul-grassi: I don't know why we need 'separate' since we are already talking about a federation.  What about 'to an RP participating in the federation'.

jimfenton: Let's just drop the word then. I'm not sure that I'd say that RPs participate in federations; they rely on them but unless you're referring to signals they aren't active participants. I will change.

sp800-63-3/cover.md
jimfenton: Good catch. I didn't apply those changes to the introduction and executive summary in -63B, so when I cut and pasted the text here, it didn't get picked up. Look for another commit to fix this and a pull request to make -63B self-consistent.

sp800-63-3/sec4_model.md
jimfenton: I hope the latter too.

paul-grassi (approved):

PR #415, opened 2016-10-09
Addresses series of editorial issues

jricher:


sp800-63c/index.md
```diff
@@ -8,38 +8,6 @@ description: "SP 800-63C"
 
 This is a working draft of NIST Special Publication 800-63C, *Assertions and Federation*. This document is a sub-document referenced by [SP 800-63-3](../sp800-63-3/) covering the associated topics that had been previously in SP 800-63-2.
```
jimfenton, line 9: I'm not sure we even reference index.md from -63a, -63b, or -63c anywhere. In -3, index.md is the "home page" for the project. So these changes are fine, but we should perhaps get rid of these files entirely (or change them to redirects to the normal views) in the longer run.

sp800-63c/sec4_federation.md
```diff
@@ -83,13 +83,22 @@ A proxied federation model can provide various benefits. For example, federation
...
-2.	The federation proxy does not blind the RP and IdP from one another. The federation proxy is able to monitor and track all subscriber relationships between the RPs and IdPs, but has no visibility into any attributes it is transmitting in the assertion.
-3.	The federation proxy blinds the RP and IdP from each other. The federation proxy is able to monitor and track all subscriber relationships between the RPs and IdPs, and has visibility into any attributes it is transmitting in the assertion.
-4.	The federation proxy blinds the RP and IdP from each other. The federation proxy is able to monitor and track all subscriber relationships between the RPs and IdPs, but has no visibility into any attributes it is transmitting in the assertion.
-5. The federation proxy blinds the RP, IdP, and itself. The federation proxy cannot monitor or track any subscriber relationships, and has no visibility into any attributes it is transmitting in the assertion. 
+<div class="text-center" markdown="1">
+
+**Table 4-1: Federation Proxies
```
jimfenton: I think you need a pair of closing asterisks here.

jimfenton (approved):
Very minor suggestions, but fine by me.

sp800-63c/sec4_federation.md
jricher: That was copied from one of the other table labels, so if that's the case then both need them.

sp800-63c/index.md
jricher: I couldn't find references either, but I saw that they still existed in -A and -B so I trimmed the content and kept it in -C.

jricher:
Addressed comments, pull when ready.

PR #419, opened 2016-10-15
Added section on key management to -C

jricher:
and clarified that assertions can be signed symmetrically or asymmetrically if the keys are managed correctly.


jimfenton (approved):
Not directly on topic for this pull request, but what is an "appropriate" algorithm? In other places, use of approved cryptography is required, and "approved" is defined. I wonder if we should be using approved rather than appropriate here.

jricher:
Yes, we should adopt consistent language.

jricher:
Changed appropriate algorithm to approved cryptography throughout.

PR #420, opened 2016-10-17
CSP-verifier security and XSS/CSRF discussion

jimfenton:


sp800-63b/sec5_authenticators.md
```diff
@@ -399,4 +399,8 @@ The latter class of verifier impersonation resistant protocols relies on access
 
 In contrast, authenticators that involve the manual entry of an authenticator output, such as out of band and one-time password authenticators, SHALL NOT be considered verifier impersonation resistant because they assume the vigilance of the claimant to determine that they are communicating with the intended verifier.
 
+#### <a name="csp-verifier"></a>5.2.6. Verifier-CSP commmunications
 
+In situations where the verifier and CSP are separate entities, communications between the verifier and CSP SHALL occur through a mutually-authenticated secure channel (such as a client-authenticated TLS connection) using approved cryptography.
```
paul-grassi: This could be construed as federation/assertions, so we either need to be clear as to why it's not, or if it is, point to -C.

sp800-63b/sec5_authenticators.md
```diff
@@ -399,4 +399,8 @@ The latter class of verifier impersonation resistant protocols relies on access
...
+
+Verification of certificates using Open Certificate Status Protocol (OCSP) [[RFC 6960]](#RFC6960) SHOULD occur through an authenticated protected channel. OCSP responses SHALL be signed as specified in the OCSP specification. The verifier SHALL check responses for validity and SHALL NOT depend upon any invalid responses received.
```
paul-grassi: This is way too prescriptive for my liking.  OCSP is just checking for status of a credential (or in this case, potentially of the servers for of the CSP/Verifier).   I would prefer something like - CSP-Verifier SHALL be mutually authenticated, server credentials SHALL be PK based.  ...  What do you think?

sp800-63b/sec8_security.md
```diff
@@ -111,3 +111,9 @@ There are several other strategies that may be applied to mitigate the threats d
...
 
 In order to maintain the integrity of the authentication factors, it is essential that it not be possible to leverage an authentication involving one factor to obtain an authenticator of a different factor. For example, a memorized secret must not be usable to obtain a new list of look-up secrets.
+
+### 8.4. Session Attacks
+
+The above discussion focuses on threats to the authentication event itself, but hijacking attacks on the session following an authentication event can have similar security impacts. The session management guidelines in [Section 7](#sec7) are essential to maintain session integrity against attacks, such as cross-site scripting (XSS). In addition, it is important to sanitize all information displayed to ensure that it does not contain executable content. These guidelines also recommend that session secrets be made inaccessible to Javascript in order to provide extra protection against exfiltration of session secrets should it be possible to inject malicious Javascript.
```
paul-grassi: Can we use a better word/phrase than sanitize?  Executable content is ok, should we ensure that static/dynamic code analysis be employed such that any unintended executed code is limited?

Do we want to call out Javascript?  Browser code?  Mobile code?  Mobile code is the fed term for things like javascript.

sp800-63b/sec8_security.md
```diff
@@ -111,3 +111,9 @@ There are several other strategies that may be applied to mitigate the threats d
...
+
+Another post-authentication threat, cross-site request forgery (CSRF), takes advantage of users' tendency to have multiple sessions active at the same time. It is important to embed and verify a session identifier into web requests to prevent the ability for a valid URL or request to be unintentially or maliciously activated by the subscriber or someone impersonating the subscriber.
```
paul-grassi: Check out my edits to CSRF.  It read funny so I tweaked.  Don't want to lose your intention though.

paul-grassi (changes requested):

sp800-63b/sec5_authenticators.md
jimfenton: What I had in mind was Figure 4-1 of -3. Assertions are used when the verifier and the relying party are separate. The situation being described here is when the CSP and verifier are separate. I'm not sure how to express that in the text (maybe refer to that diagram as I just did).

sp800-63b/sec5_authenticators.md
jimfenton: The thing that's different about OCSP is that it's signed by the CSP, which means that it doesn't require a mutually authenticated channel (perhaps doesn't require an authenticated channel at all). I think it would be more prescriptive to require that all server credentials be PK based...the purpose of the mutually authenticated protected channel is to allow the CSP and verifier to "just talk"...we don't need anything too fancy here because there is (usually) a 1:1 relationship between CSPs and verifiers. A lot of the complexity of assertions is because the verifier is probably serving many RPs.

sp800-63b/sec8_security.md
jimfenton: _Sanitize_ seems to be the term of art used for making sure that user submitted content can't be used for XSS...for example see [this](https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet#RULE_.236_-_Sanitize_HTML_Markup_with_a_Library_Designed_for_the_Job).

XSS attacks are all about Javascript currently, but yes, should broaden. I prefer "browser executable code" because it's not limited to mobile devices.

Look for a push to this branch to update this.

sp800-63b/sec5_authenticators.md
paul-grassi: crap.  you're right.  A reference or the explicit text you just wrote in this discussion could be helpful  But that was also a brain fart moment for me.

sp800-63b/sec5_authenticators.md
paul-grassi: Hmm, given comment 1, i see your point.  But this is a very weird corner case, and not sure it's one that demands OCSP.  Consider something, i won't hold up the pull, but it doesn't feel all the way right to me.

sp800-63b/sec8_security.md
paul-grassi: Hah!  Yeah, mobile code isn't mobile phones.  Mobile == downloaded and executed on browser. [Mobile Code](https://www.techopedia.com/definition/10294/mobile-code).  Given feds, I would use mobile code.

paul-grassi: Point to OWAST then as an example.  Feds love OWASP.

sp800-63b/sec8_security.md
jimfenton: Yes, it did read funny; thanks. I spent enough time tweaking the text that I wasn't seeing that.

One aspect of your change I might question was where you changed:

> to be unintentially activated by the subscriber

to

> to be unintentially or maliciously activated by the subscriber or someone impersonating the subscriber

My understanding of CSRF is that it's directly or indirectly activated by the subscriber clicking on a link while in possession of a secret for the authenticated session. I don't believe that an impostor can take advantage of a CSRF-vulnerable link directly, although maybe there is some way that can be done with browser-executed code.

I'd like to get this right, but at the same time keep it kind of brief because we're getting out of the authentication topic with CSRF. If we were giving more general guidelines on keeping sessions secure, there would be a _lot_ more to cover.

paul-grassi: Oof,a fair point.  The browser guys are looking at 'containers' such that a user can have tabs open under multiple personas, without this occurring.  https://wiki.mozilla.org/Security/Contextual_Identity_Project/Containers

I like my edits because we can't think what's not possible today won't be tomorrow.  You can get rid of the subjects and just say 'unintentionally or maliciously executed'.

sp800-63b/sec5_authenticators.md
jimfenton: What it said about OCSP needing a different sort of protected channel for OCSP was a bit of a non-sequitur anyway, so I'm going to leave that off for now at least. I'm not convinced it makes sense to special case it.

sp800-63b/sec8_security.md
jimfenton: Adding the reference to OWASP and using _mobile code_ instead of JavaScript. Also adding definition of mobile code.

jimfenton: A little bit nervous about the OWASP reference because it is hosted on a wiki, but this is informative in any case.

paul-grassi (approved):

PR #424, opened 2016-10-26
Minor edits to assertion presentation based on discussion with Naomi …

paul-grassi:
…and Jamie.


sp800-63c/sec6_presentation.md
```diff
@@ -6,6 +6,8 @@ Assertions MAY be presented in either a *back-channel* or *front-channel* manner
 
 The IdP SHALL transmit only those attributes that were explicitly requested by the RP. RPs SHALL conduct a privacy risk assessment when determining which attributes to request. 
 
+To mitigate the risk of unauthorized exposure of sensitive information (e.g., shoulder surfing), the CSP SHALL, by default, mask sensitive information displayed to the subscriber.  The CSP SHALL provide mechanisms for the subscriber to temporarily unmask such information in order for the subscriber to view full values.
```
jricher, line 9: We don't have a definition of what it means to "mask" information. For example, we should allow the CSP to "mask" by not displaying the information in the first place, and the current language choice sounds like it's only black-out.

jricher (changes requested):

sp800-63c/sec6_presentation.md
paul-grassi: I'll add something. We also have the UX chapter coming, that will have more language about masking.

jricher (approved):

PR #425, opened 2016-10-26
Final fentonpalooza2 updates

paul-grassi:
@jimfenton any help on claimed identity would be great.  I don't think we should delete per the issue, but I attempted to get rid of the name, dob, address reference.

Also, in the spirit of closing issues, I did a minor fix to -3, knowing much of the current text may be OBE when I get the big edits in there.  In other words, don't kill yourself on the 2nd commit :)


sp800-63-3/sec4_model.md
```diff
@@ -182,6 +182,7 @@ Agency mission need will assist in determining the acceptable IAL at a given LOA
...
+> Note: An agency can accept a higher assurance level than those required in the table above.  For example, in a federated transaction, an agency can accept an IAL3 identity if their application is assessed at IAL2.  The same holds true for authenticators; stronger authenticators can be used at RP's that have lower authenticator requirements.  However, RPs will ensure that these scenarios only occur in federated scenarios with appropriate privacy protections by the CSP to ensure that only the requested attributes are provided to the RP and that no personal information leaks from the authenticator or the assertion.  See [privacy requirements](./sp800-63c.html#sec9) in SP 800-63C for more details.
```
jimfenton, line 185: Nit: RP's -> RPs

sp800-63-3/sec4_model.md
```diff
@@ -182,6 +182,7 @@ Agency mission need will assist in determining the acceptable IAL at a given LOA
...
 
-As agencies are encouraged to consider each distinct element of assurance, the notion of the 'low watermark' to determine LOA no longer applies.  An IAL1/AAL2 application should not be considered any less secure than an IAL2/AAL application.  The only difference between these applications is the amount of proofing required, which does not impact security of each application. That said, if an agency incorrectly determines the xAL, security could very well be impacted.
\ No newline at end of file
+
+> Note: Agencies are encouraged to consider each distinct element of assurance, therefore the notion of the 'low watermark' to determine LOA no longer applies.  An IAL1/AAL2 application should not be considered any less secure or privacy enhancing than an IAL2/AAL2 application.  The only difference between these applications is the amount of proofing required, which may not impact the security and privacy of each application. That said, if an agency incorrectly determines the xAL, security and privacy could very well be impacted.
```
jimfenton, line 188: Since this is the only mention of _low watermark_ in the document, wonder if there is enough context for what this is talking about. Should this be introduced like "In earlier versions of this Guideline, the concept of 'low watermark' was used..." or something like that?

sp800-63-3/sec3_definitions.md
```diff
@@ -62,7 +62,7 @@ In this document, biometrics may be used to unlock authenticators and prevent re
 A party whose identity is to be verified using one or more authentication protocols.
 
 #### Claimed Identity
-A declaration by the applicant of their current Personal Name, date of birth and address. [[GPG45]](#gpg45)
+A declaration of unvalidated and unverified personal attributes by the applicant.
```
jimfenton, line 65: This definition should also appear in -63A, where the term is used extensively. There is currently no definition for _claimed identity_ there.

jimfenton (changes requested):
Adding the definition to -63A is the most important of the comments.

paul-grassi:
ok, on a call now, but will fix around 4est.

paul-grassi:
@jimfenton done.  Fix came earlier than expected :)

jimfenton (approved):

PR #427, opened 2016-11-02
Small edits in response to internal comments

jimfenton:


paul-grassi (approved):
I'm really not in love with 'as of the date of this publication', but until we can come up with something better, I'll merge.

PR #436, opened 2016-11-21
X al cyoa

paul-grassi:
@jimfenton here is the pull we discussed.  And great, can't auto-merge.  We can discuss this afternoon.  I want to get this up there but it's not close to final yet.  Will get that finished this week.


jimfenton (approved):
Yup.

paul-grassi:
Thanks.  Looks like 2 conflicts thankfully.  Can you help with that?

PR #450, opened 2016-11-28
Minor edits before finalizing the exec summary.

paul-grassi:


sp800-63-3/sec4_model.md
```diff
@@ -81,7 +81,7 @@ The secrets contained in authenticators are based on either public key pairs (as
...
+In this document, authenticators always contain a secret. Some of the classic authentication factors do not apply directly to digital authentication. For example, an ID badge is something you have, and is useful when authenticating to a human (e.g., a guard), but is not a authenticator for digital authentication. Authentication factors classified as something you know are not necessarily secrets, either. Knowledge based authentication, where the claimant is prompted to answer questions that can be confirmed from public databases, also does not constitute an acceptable secret for digital authentication. More generally, something you are does not generally constitute a secret. Accordingly, this guideline only allows the use of biometrics as as part of a multifactor authenticator.
```
jimfenton: I'm a little concerned that the new sentence might be misinterpreted as applying outside authentication. The guideline does allow the use of biometrics in conjunction with registration, which might be considered to contradict this. How about something like, "Accordingly, this guideline only allows the use of biometrics in authentication only as an activation factor for a specific physical authenticator."

jimfenton (changes requested):

paul-grassi:
We are nitpicking either way. The old language of not allowed looked awkward. You're way is fine for now, but doesn't protect against innovation and change in approach. Granted this is informative language but it should be right. so I prefer to stay with the language I have but won't fight hard about it.

jimfenton:
How about a gentler rewording, like, "Accordingly, this guideline only allows the use of biometrics for authentication as part of a multifactor authenticator."

In any case, need to fix the "as as".

jimfenton (approved):

PR #454, opened 2016-12-02
Modifed intro to align with rest of document

paul-grassi:


jimfenton (approved):
Looks good. I was wondering a little about the deletion of the text about KBA; I wonder if it would be helpful to mention that KBA isn't allowed for authentication like it had.

PR #455, opened 2016-12-02
Misc cleanup.

paul-grassi:
@jimfenton see if you are ok with the enroll code time to live for IAL3 in-person.


jimfenton:
Yes, 7 days seems like a very reasonable time limit.

jimfenton (approved):

PR #461, opened 2016-12-06
Adjustments to IAL/AAL compatibility matrix

jimfenton:


paul-grassi (approved):
@jimfenton will pull in now.  I do think we should do something more for IAL1/AAL1, but I will put some thoughts into the table and shoot it your way.

PR #462, opened 2016-12-08
Update IAL/AAL combo table

paul-grassi:
@jimfenton, I felt we had to call out when personal data is captured or not, per our discussion (especially the use case I described).  In addition, I used the EO language 'personal data', not PII, just to be consistent.

### DIFF
--- a/sp800-63-3/sec5_xAL.md
+++ b/sp800-63-3/sec5_xAL.md
@@ -93,7 +93,7 @@ The ability to combine varying xALs offers significant flexibility to agencies, 
 | **IAL2** |  **NO** | Allowed | Allowed |
 | **IAL3** |  **NO** | Allowed | Allowed |
 
-> Note: Per Executive Order 13681 [[EO 13681]](#EO13681), the release of personal data requires protection with multifactor authentication, even if the personal data is self-asserted and not validated. However, depending on the use case, in particular when the transaction that the user authenticates to does not make personal data accessible, it could be possible for authentication to occur at AAL 1, though unlikely and not recommended. In addition, it may be possible at IAL1 to self-assert information that is not personal, in which case AAL1 is acceptable.
+> Note: Per Executive Order 13681 [[EO 13681]](#EO13681), the release of personal data requires protection with multifactor authentication, even if the personal data is self-asserted and not validated. When the transaction does not make personal data accessible, authentication may occur at AAL 1, although stronger authentication is recommended. In addition, it may be possible at IAL1 to self-assert information that is not personal, in which case AAL1 is acceptable.
 
  
 ### 5.3. <a name="CYOA"></a>Selecting the Appropriate xAL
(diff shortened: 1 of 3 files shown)


sp800-63-3/sec5_xAL.md
```diff
@@ -86,13 +86,14 @@ The ability to combine varying xALs offers significant flexibility to agencies,
...
+| **IAL2** |  **NO** | Allowed | Allowed |
+| **IAL3** |  **NO** | Allowed | Allowed |
 
-> Note: Per Executive Order 13681 [[EO 13681]](#EO13681), the release of personal information requires protection with multifactor authentication. Therefore, it should not be possible to use single-factor authentication (AAL 1) to release personally identifiable information from a CSP. However, depending on the use case, in particular when the transaction that the user authenticates to does not release or manage any personal information, it could be possible for authentication to occur at AAL 1, though unlikely and not recommended.
+> Note: Per Executive Order 13681 [[EO 13681]](#EO13681), the release of personal data requires protection with multifactor authentication, even if the personal data is self-asserted and not validated. However, depending on the use case, in particular when the transaction that the user authenticates to does not make personal data accessible, it could be possible for authentication to occur at AAL 1, though unlikely and not recommended. In addition, it may be possible at IAL1 to self-assert information that is not personal, in which case AAL1 is acceptable.
```
jimfenton: OK, but seems a little wordy. For second sentence, suggest something like: "When the transaction does not make personal data accessible, authentication may occur at AAL 1, although stronger authentication is recommended."

jimfenton (approved):
Agree with using the EO 13681 terminology, and allowing non-personal data at AAL 1. However, this seems to be establishing a normative requirement and the -3 volume is only informative. I don't know where else to put this, but maybe we need to reconsider whether -3 is informative in its entirety, because it's the logical place to put guidelines like this that tie -63A, -63B, and -63C together.

sp800-63-3/sec5_xAL.md
paul-grassi: That works for me.  I can make the change.

paul-grassi:
I really want to try keeping it informative.  In the beginning of norm requirements in -A, -B, or -C we say what is allowable.  Since M-04-04 is the NORMATIVE doc in this case, I am hesitant to make anyone do anything.

paul-grassi:
@jimfenton changes made, and normative in -A/B aligned.

sp800-63b/sec4_aal.md
```diff
@@ -38,11 +38,13 @@ However, [Table 4-2](#63bSec4-Table2) shows the expanded set of AALs that are al
...
 | M-04-04 Level of Assurance | Authenticator Assurance Level
 |:------------------:|:-----------------------------:
-| 1 | 1, 2 or 3 
+| 1; without making personal data available | 1, 2 or 3 
+| 1; making personal data available | 2 or 3
 | 2 | 2 or 3
 | 3 | 2 or 3 
 | 4 | 3 
 
+At IAL1, it is possible that self-asserted personal data is collected and made available by the digital service.  Any personal data, self-asserted or validated, requires MFA; therefore agencies SHALL select a minimum of AAL2 when self-asserted personal data, collected at IAL1, is made available online.
```
jimfenton: -63B does not already use the abbreviation MFA anywhere, so this should be spelled out.

PR #464, opened 2016-12-15
edits based on @lrl comments.

paul-grassi:
@lrl had some nice comments that went into this edit. I'm done after this as @garciamike is starting to review


sp800-63-3/sec5_xAL.md
```diff
@@ -110,9 +112,11 @@ Separating IAL and AAL also works for traditional M-04-04 LOA use cases, such as
...
+An online transaction may not be equivalent to a complete business process that requires offline processing, or online processing in a completely separate system. The agency should assess the risk of the portion of a business process available via the digital service, not the entire business process that is executed outside the digital service to provide the benefit or service. For example, in an online survey, sensitive information may be collected, but it is never made available online after the information is submitted. In this instance, it is important for the information to be carefully protected in backend systems, but there is no reason to identity proof or even authenticate the user providing the information. The online transaction is solely a submission of the data only. The entire business process may require a significant amount of data validation, but to submit the data there is no need proof or authenticate.
```
jimfenton: Nit: "need proof or authenticate" -> "need to proof or authenticate"

sp800-63-3/sec5_xAL.md
```diff
@@ -110,9 +112,11 @@ Separating IAL and AAL also works for traditional M-04-04 LOA use cases, such as
...
+
+Another example is a digital service that accepts resumes to apply for open job postings. In this case, the service allows any individual to submit a resume on behalf of someone else, and in subsequent visits to the site, access the resume. Since the resume contains personal information and is provided in later sessions, the agency must select an AAL2 authenticator, at a minimum.
```
jimfenton: Agencies don't select the authenticator, they select the AAL. So "agency must select an AAL2 authenticator" -> "agency must select at least AAL2 authentication" (is it "must" or "shall"?)

sp800-63-3/sec5_xAL.md
```diff
@@ -110,9 +112,11 @@ Separating IAL and AAL also works for traditional M-04-04 LOA use cases, such as
...
+However, the identity proofing requirements may not require the same rigor. The entire business process of examining a resume and interviewing a person requires a significant amount of identity proofing. The agency needs a high level of confidence that the job applicant is in fact the subject of the resume submitted online. Yet proofing is not required to submit the resume. Identity proofing is not required to complete the digital transaction successfully. The resume could be submitted by the applicant, but may also be submitted by a surrogate. Identity proofing the submitter would create more risk than required as personal information would be collected for the surrogate before she could submit any resumes. Therefore, the most appropriate IAL selection would be 1.
```
jimfenton: I'm not sure this should be a separate paragraph, since it refers to the same use example.

The middle sentences seem a little choppy and redundant, suggest instead:

> Yet proofing is not required to submit the resume, since it could be submitted by either the applicant or a surrogate such as a recuiter. Requiring identity proofing of the submitter would create more risk than required as personal information would be required regarding the surrogate before they could submit any resumes. Therefore, the most appropriate IAL selection would be 1.

jimfenton (changes requested):
A few nits and wording clarifications

sp800-63-3/sec5_xAL.md
paul-grassi: well they do select the authenticators too, but in the spirit of federation, i will make this change :)

paul-grassi:
@jimfenton new commit added.

jimfenton (approved):
Looks good, thanks for changes.

PR #465, opened 2016-12-20
Final priv edits

paul-grassi:


sp800-63c/sec9_privacy.md
```diff
@@ -33,9 +32,9 @@ Due to the many components of digital authentication, it is important for the SA
...
 
-Note that even with the use of blinding technologies, it MAY still be possible for a blinded party to infer protected subscriber information through released attribute data or metadata such as analysis of timestamps, attribute bundle sizes, or attribute signer information. 
+It may still be possible for a blinded party to infer protected subscriber information through released attribute data or metadata such as analysis of timestamps, attribute bundle sizes, or attribute signer information. The IdP could consider additional privacy-enhancing approaches to reduce the risk of revealing identifying information of the entities participating in the federation.
```
jricher: We should restore the "Even with the use of blinding technologies, it may still..." as this is the hinge of the caveat.

paul-grassi: This is in the Blinding Section so I felt that caveat was redundant.  I don't feel strong enough to debate this one, so if you still think it should be there, I can put it back.

nbl-nist (approved):

nbl-nist:
LGTM

sp800-63c/sec9_privacy.md
jricher: I think that it helps connect the thoughts together, so I'd like to see it back in. I agree with the removal of the normative MAY (I think that was an editing mistake in the first place).

paul-grassi: Just committed the change.

jricher (approved):

PR #468, opened 2016-12-31
Updates from Mike's -63B review

jimfenton:
This includes the changes from Mike's 63a-pre-public-comment branch to -63B, so please don't merge those changes because that would cause confusion.


sp800-63-3/cover.md
```diff
@@ -91,14 +91,14 @@ http://dx.doi.org/10.6028/NIST.SP.XXX
 
 {::comment}
 
-**Comments on this publication may be submitted to eauth-comments@nist.gov  
+**Comments on this publication may be submitted to eauth-comment@nist.gov
```
garciamike: Do we want a new email address? We can forward the eauth-comment addy if we want to retain that one.

paul-grassi: I want nothing more.

garciamike: @kboeckl, can you get a new email address? I don't know if this one is an alias or a full inbox. whatever address paul fancies. thanks!

jimfenton: The change was made because we were inconsistent; eauth-comments@ was used on previous revisions of 800-63 and eauth-comment@ was used for the latest NTR. If you want to change it to something more consistent with the current title, we can do that too, but let's make them consistent.

sp800-63b/cover.md
```diff
@@ -3,7 +3,7 @@
 
 # <a name="800-63b"></a> DRAFT NIST Special Publication 800-63B
 
-# Digital Authentication Guideline
+# Digital Identity Guidelines
 
 ### Authentication and Lifecycle Management
```
paul-grassi, line 8: Since we're in the business of renaming, any reason 'Authenticator' wouldn't make the cut?  'Authenticators and Lifecycle management'?  I don't feel strongly.

sp800-63b/cover.md
```diff
@@ -184,8 +186,7 @@ activities with industry, government, and academic organizations.
...
-of the three authenticator assurance levels. This publication supersedes corresponding sections of NIST SP 800-63-1 and SP 800-63-2.
+This document and its companion documents, SP 800-63-3, SP 800-63A, and SP 800-63C, provide technical and procedural guidelines to agencies implementing electronic authentication to choose and implement effective authentication processes based on risk. The recommendation covers remote authentication of users (such as employees, contractors, or private individuals) interacting with government systems over open networks. It defines technical requirements for each of the three authenticator assurance levels. This publication supersedes corresponding sections of NIST SP 800-63-1 and SP 800-63-2.
```
paul-grassi: Check out what I did in -A for this section and see what you think.  I like what I did, but happy to back out.  I like that it was specific to -A, not repeated and generic.

sp800-63b/sec1_2_introduction.md
```diff
@@ -2,24 +2,28 @@
...
+_This section is informative._
+
+This document and its companion documents, [SP 800-63-3](sp800-63-3.html), [SP 800-63A](sp800-63a.html), and [SP 800-63C](sp800-63c.html), provide technical guidelines to agencies for the implementation of digital authentication.
 
 <a name="sec2"></a>
 
 ## 2. Introduction
 
-Digital authentication is the process of establishing confidence in user identities electronically presented to an information system. E-authentication presents a technical challenge when this process involves the digital authentication of individual people over a network. 
+_This section is informative_
```
paul-grassi: Need a period at the end of *This section is informative*

sp800-63b/sec4_aal.md
```diff
@@ -82,15 +84,15 @@ The CSP SHOULD employ appropriately tailored security controls from the low base
...
 
-AAL 2 provides high confidence that the claimant controls the authenticator registered to a subscriber. Two different authentication factors are required. Approved cryptographic techniques are required at AAL 2 and above.
+AAL 2 provides high confidence that the claimant controls authenticator(s) registered to the subscriber. Proof of possession and control of two different authentication factors is required through a secure authentication protocol. Approved cryptographic techniques are required at AAL 2 and above.
 
 #### 4.2.1. Permitted Authenticator Types
 
-At AAL 2, it is required to have (a) a multi-factor authenticator, or (b) a combination of two single-factor authenticators. Authenticator requirements are specified in Section 5.
+At AAL 2, it is required to have either a multi-factor authenticator or a combination of two single-factor authenticators. Authenticator requirements are specified in Section 5.
 
 When a multi-factor authenticator is used, any of the following may be used:
```
paul-grassi, line 97: Not a showstopper for Kevin, but we are going to need a hearty reason why 2 SF authenticators can't include a biometric - per our discussion last week with Elaine.

sp800-63b/sec4_aal.md
```diff
@@ -192,10 +196,8 @@ Requirement | AAL 1 | AAL 2 | AAL 3
 **FIPS 140 verification** | Level 1 (Government agency verifiers) | Level 1 (Government agency authenticators and verifiers) | Level 2 overall (MF authenticators)<br />Level 1 overall (verifiers and SF Crypto Devices)<br />Level 3 physical security (all authenticators)
 **Reauthentication** | 30 days | 12 hours or 30 minutes inactivity; may use one authentication factor | 12 hours or 15 minutes inactivity; shall use both authentication factors
 **Security controls**|[[SP 800-53]](#SP800-53) Low Baseline (or equivalent)|[[SP 800-53]](#SP800-53) Moderate Baseline (or equivalent)|[[SP 800-53]](#SP800-53) High Baseline (or equivalent)
-**Privacy controls**|[[SP 800-53]](#SP800-53)(or equivalent)|[[SP 800-53]](#SP800-53) (or equivalent)|[[SP 800-53]](#SP800-53) (or equivalent)
 **MITM resistance** | Required | Required | Required |
 **Verifier impersonation resistance** | Not required | Not required | Required |
```
paul-grassi, line 199: What happened to VCR?  I thought we were going to throw it out there with an NTR?

sp800-63b/sec5_authenticators.md
```diff
@@ -31,23 +33,23 @@ Memorized secrets that are randomly chosen by the CSP (e.g., at enrollment) or b
 
 Memorized secret verifiers SHALL NOT permit the subscriber to store a "hint" that is accessible to an unauthenticated claimant. Verifiers also SHALL NOT prompt subscribers to use specific types of information (e.g., "What was the name of your first pet?") when choosing memorized secrets.
 
-When processing requests to establish and change memorized secrets, verifiers SHALL compare the prospective secrets against a list of known commonly-used, expected, and/or compromised values. For example, the list MAY include (but is not limited to):
+When processing requests to establish and change memorized secrets, verifiers SHALL compare the prospective secrets against a list that contains values known to be commonly-used, expected, or compromised. For example, the list MAY include (but is not limited to):
 
 * Passwords obtained from previous breach corpuses
```
paul-grassi, line 38: @garciamike i think this needs to remain, but this is against policy and perhaps the law.  Do we need to qualify it?

sp800-63b/sec5_authenticators.md
```diff
@@ -102,11 +104,11 @@ The out-of-band device SHALL be uniquely addressable and communication over the
 
 The out-of-band authenticator SHALL uniquely authenticate itself in one of the following ways in communicating with the verifier:
 
-- Establish an authenticated protected channel to the verifier using approved cryptography. The key used SHALL be stored in the most secure storage available on the device (e.g., keychain storage, trusted platform module, or trusted execution environment if available).
+- Establish an authenticated protected channel to the verifier using approved cryptography. The key used SHALL be stored in the most secure storage available on the device (e.g., keychain storage, trusted platform module, trusted execution).
```
paul-grassi: TEE is a common term, why did we drop 'environment'?

sp800-63b/sec5_authenticators.md
```diff
@@ -314,21 +316,19 @@ CSPs SHALL provide subscriber instructions on how to appropriately  protect the
 
 #### <a name="throttle"></a>5.2.2. Rate Limiting (Throttling)
 
-*cf. 800-63-2 sec 8.2.3, p.75*
-
-When the authenticator output or activation secret does not have sufficient entropy, the verifier SHALL implement controls to protect against online guessing attacks. Unless otherwise specified in the description of a given authenticator, the verifier SHALL effectively limit online attackers to 100 consecutive failed attempts on a single account in any 30 day period.
+When the authenticator output or activation secret does not have sufficient entropy, the verifier SHALL implement controls to protect against online guessing attacks. Unless otherwise specified in the description of a given authenticator, the verifier SHALL effectively limit online attackers to 100 consecutive failed attempts on a single account.
```
paul-grassi: I'm not sure I would restrict this if the authenticator doesn't have enough entropy.  A 'strong' password should still be rate limited. I have also never been happy with 100 attempts.  Industry standard is much less, can we drop this down?  If velocity is an issue, we can make that clear.

sp800-63b/sec5_authenticators.md
```diff
@@ -314,21 +316,19 @@ CSPs SHALL provide subscriber instructions on how to appropriately  protect the
...
 
-- Requiring the claimant to wait for a short period of time (anything from 30 seconds to an hour, depending on how close the system is to its maximum allowance for failed attempts) before attempting Authentication following a failed attempt
+- Requiring the claimant to wait following a failed attempt for a period of time that is increasing in intervals from, say, 30 seconds to an hour, as the account approaches its maximum allowance for consecutive failed attempts
 
 - Only accepting authentication requests from a white list of IP addresses at which the subscriber has been successfully authenticated before
 
 - Leveraging other risk-based or adaptive authentication techniques to identify user behavior that falls within, or out of, typical norms.
 
-Since these measures often create user inconvenience, the verifier SHOULD allow a certain number of failed authentication attempts before employing the above techniques.
+Since these measures often create user inconvenience, the verifier SHOULD allow a several failed authentication attempts before employing the above techniques.
```
paul-grassi: I think we should delete this requirement.  Make this a UX consideration. If service wants to lock out after 1 attempt, so be it.  'Since these measures often create user inconvenience, the verifier SHOULD allow a several failed authentication attempts before employing the above techniques.'

sp800-63b/sec5_authenticators.md
```diff
@@ -359,7 +359,7 @@ Biometric matching SHOULD be performed locally on claimant's device or MAY be pe
 
 If matching is performed centrally:
 
-* Use of the biometric SHALL be bound tightly to a single, specific device that is identified using approved cryptography.
+* Use of the biometric SHALL be limited to a single, specific device that is identified using approved cryptography.
```
paul-grassi: This isn't what we mean.  If my fingerprint is on the server, and the agency wants to issue me a laptop, ipad, iphone, samsung, they should - each being trusted and revocable.

paul-grassi (changes requested):
@jimfenton a few things here that I would like clarification on before we merge in for Kevin.  Other that, looks awesome.

sp800-63b/cover.md
jimfenton: Authenticators are just a piece of this (the part kept by the subscriber), and we're talking about the whole process here (including verifiers, etc.) so I think that might be too narrow.

sp800-63b/cover.md
jimfenton: Yes, I was looking at the -A abstract this morning. Let me try changing it to something in that style.

sp800-63b/sec4_aal.md
jimfenton: But it can -- the biometric has to be bound to a single specific device that is identified using approved cryptography (sec. 5.2.3). I interpret that as requiring that the biometric be presented along with a specific second authenticator -- not just any authenticator associated with the user and not, for example, a memorized secret.  Maybe I was reading this more narrowly last week.

sp800-63b/sec4_aal.md
jimfenton: I have the VCR text sitting out in another branch (jimfenton/b-verifier-secrets). I didn't think we had consensus from the other folks we consulted to include it, but I can do that any time.

sp800-63b/sec5_authenticators.md
jimfenton: I'm not sure; I'll put it back. @garciamike please speak up if we're missing the point of your edit.

sp800-63b/sec5_authenticators.md
jimfenton: Agreed.

sp800-63b/sec5_authenticators.md
jimfenton: I thought we were specifying a maximum number of failed attempts, but it reads like an absolute number. How about if we say "at most 100 consecutive failed attempts" and allow verifiers to profile downward if they want?

sp800-63b/sec5_authenticators.md
jimfenton: I'm not sure that works with the next bullet, which requires use of templates and biometric revocation. The idea is that you enroll and use the biometric with a specific authenticator -- the server should maintain a separate template for each device, somewhat analogous to a hash of a password. But might want to confirm with @enewt .

sp800-63b/cover.md
paul-grassi: Cool!

sp800-63b/sec4_aal.md
paul-grassi: It's this requirements I am talking about:

> When a combination of two single-factor authenticators is used, it SHALL include a Memorized Secret authenticator and one possession-based (“something you have”) authenticator from the following list:

>Look-up Secret
Out of Band
Single Factor OTP Device
Single Factor Cryptographic Software
Single Factor Cryptographic Device

I remain unconvinced that this is the right thing to say, and that biometrics can only unlock a secret.  But I am willing to stick with it for the time being (knowing we will get killed in public comment).  So, what I am asking for is more explanation in the text as to why we are making this statement.  If it was up to me, this entire part would be dropped and industry would figure out a way to deploy authenticators in any combination they want, ensuring if it is a biometric it meets our security and performance requirements.  And revocability with MF authenticators is not a good reason.

sp800-63b/sec4_aal.md
paul-grassi: Well, I never loved the added requirement, I just liked the informational aspect of what authenticators have it, and which ones don't.  I would prefer that at a minimum to go into the public comment.  But if you want to stick with the requirement, then send a PR and lets have a big-fat NTR for this one :) (per Bill's suggestion).

sp800-63b/sec5_authenticators.md
paul-grassi: The whole point of server side is to enroll once, use many.  Otherwise, I would just enroll per device and use the local biometric.  I will talk to @enewt at 11 and give you a call after, but I think the requirement needs to change.

sp800-63b/sec5_authenticators.md
paul-grassi: Speaking with him at 11, will let you know.

sp800-63b/sec5_authenticators.md
paul-grassi: Well, we lost the 30 days too for some reason, so I don't know what the heck is going on with this.  I thought we were all good with 30 days and 'at most...'.  I would like that back in.  I think that means rejecting Mike's edit.

sp800-63b/sec5_authenticators.md
paul-grassi: Yay!

sp800-63b/sec5_authenticators.md
paul-grassi: Confirmed with @enewt that where  I was heading is correct.  Recommend dropping the second bullet and changing the 1st to:  Use of the biometric SHALL be bound tightly to a trusted device that is identified using approved cryptography.

sp800-63b/sec5_authenticators.md
paul-grassi: Confirmed with @garciamike TEE should remain.

sp800-63b/sec4_aal.md
jimfenton: Having it be just informational isn't actionable by anyone. I'd like to throw it in as a requirement because that is what will prompt feedback. But let's merge this first and I'll rebase that change on top of this.

sp800-63b/sec5_authenticators.md
jimfenton: Do we need to define "trusted device"? Trusted by the subject, or by the verifier/CSP? My concern is that there is no standard basis for establishing trust, and nothing to prevent the verifier/CSP from deciding to issue blanket trust to whole classes of sensors that could be compromised.

sp800-63b/sec4_aal.md
paul-grassi: 👍 

That said, agencies will still be picking authenticators.  Letting them know which are better helps.

sp800-63b/sec5_authenticators.md
garciamike: I'm okay with it as it is but if we get any pushback we can add some legalistic language.

sp800-63b/sec5_authenticators.md
paul-grassi: Fine, then drop 'single'.  That is misleading.  I can have multiple trusted devices using a single enrolled template on a server.

sp800-63b/sec4_aal.md
```diff
@@ -108,7 +108,7 @@ When a combination of two single-factor authenticators is used, it SHALL include
 * Single Factor Cryptographic Software
 * Single Factor Cryptographic Device
 
-> Note: The requirement for a memorized secret authenticator above derives from the need for two different types of authentication factors to be used. All biometric authenticators compliant with this specification are multi-factor, so something you know (a memorized secret) is the remaining possibility.
+> Note: When biometric authentication is used, use of the biometric SHALL be limited to specific devices as described in Section [5.2.3](#biometric_use). Therefore, the device is "something you have" which serves as a second factor for the authentication. Another factor such as a memorized secret is not required (but MAY be used to provide additional assurance).
```
paul-grassi, line 111: Let's keep working this, but for now, we are good.  I don't like the SHALL in  a note, but again, we can fix later.

sp800-63b/sec5_authenticators.md
```diff
@@ -393,6 +393,12 @@ The latter class of verifier impersonation resistant protocols relies on access
...
+
+#### <a name="replay"></a>5.2.7. Replay resistance
+
+Authenticators whose output is usable only once are referred to as replay resistant. Replay resistance is in addition to the replay resistant nature of authenticated protected channel protocols, since the output could be stolen prior to entry into the protected channel. Examples of replay resistant authenticators are OTP devices, cryptographic authenticators, and look-up secrets.
+
+In contrast, memorized secrets SHALL NOT be considered replay resistant because the authenticator output (the secret itself) is provided for each authentication. Although not recognized as an independent authenticator, biometrics SHALL NOT be considered replay resistant because a match result can often be obtained if the sensor output is recorded and a subtly manipulated version is replayed, even though the replay may not have been exact.
```
paul-grassi, line 404: Same.  I think biometrics can be replay resistant.  It's not up to us to agree on what is and isn't possible, let innovation occur.  I think we should discuss, I'm not comfortable giving this to Kevin as-is.

That said, thanks for everything, it's looking awesome, and that is largely thanks to you!

paul-grassi (approved):

PR #471, opened 2017-01-06
Added a new dictionary per Kelley D and fixed a grammar edit.

paul-grassi:


jimfenton (approved):
Looks good. Holding off on merge for clarification of wording on max allowable length of memorized secrets.

jimfenton:
I like your wording better than mine. WRT the "repetitive or sequential" part, this is fine since it's an example but presumably common ones would be part of a commonly-used passwords dictionary anyway. There are also typing patterns (qwerty, qazwsxedc etc.) and leetspeak transformations (p4ssw0rd) but it's probably counterproductive to try to enumerate all of the things that might be included.

readnotify:
Password requirements when a user has MFA turned on are worse than counter productive: i.e. how is a modern cellphone user supposed to recover access to 101+ accounts the day they loose their phone, if your rules have forced them to create 101 un-rememberable chunks of gibberish, when they in fact needed none in the first place owing to the user already having good MFA activated on them all?

Most passwords are not compromised in ways any of these rules will resist - most passwords are stolen.  These rules do nothing for security, but ruin user lives.

PR #472, opened 2017-01-07
Updates from Mike's -63C review

jimfenton:


sp800-63c/sec10_usability.md
```diff
@@ -36,12 +37,12 @@ This section is organized by the functional elements a user may encounter when i
...
+Federated identity systems typically redirect the user from the RP to the IDP, and back again. This could be jarring to users for various reasons.  They may think the redirects are the result of a malicious entity.  Or they may be concerned about leaving one domain for another, especially if the root domain changes (e.g., .gov to .com).  In addition, RPs may be concerned with branding and any friction that may keep a user from executing their intended transaction.  The following list considers tactics that can be employed to resolve negative user experience; in addition, these techniques could facilitate RP acceptance of temporarily turning control of the user experience to the IdP as the authentication request proceeds. Possible considerations include:
```
jricher: IDP -> IdP

sp800-63c/sec5_assertions.md
```diff
@@ -81,11 +81,11 @@ In some circumstances, it is desirable to prevent the subscriber's account at th
...
 
 #### 5.2.6. <a name="ppi-gen"></a>Pairwise Pseudonymous Identifier Generation
 
-Pairwise pseudonymous identifiers SHALL be opaque and unguessable, containing no identifying information about the subscriber. Normally, the identifiers SHALL only be known by and used by one IdP-RP pair. However, an IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if:
+Pairwise pseudonymous identifiers SHALL be opaque and unguessable, neither containing nor generated from identifying information about the subscriber. Normally, the identifiers SHALL only be known by and used by one IdP-RP pair or IdP-Proxy and Proxy-RP pair. However, an IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if:
```
jricher: Reject these edits as previously discussed. Change to:

> Pairwise pseudonymous identifiers SHALL be opaque and unguessable, containing no identifying information about the subscriber. Pairwise pseudonymous identifiers SHOULD be random values with sufficient entropy to prohibit guessing, and SHALL NOT be generated from identifying information in such a way as to allow an attacker to derive the identifying information from the identifer alone (such as through a dictionary attack on a hashed value). Normally, the identifiers SHALL only be known by and used by one IdP-RP pair. However, an IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if:

sp800-63c/sec9_privacy.md
```diff
@@ -4,37 +4,41 @@
...
 
-Section 4.2 also encourages the use of technical measures to provide unlinkability and prevent subscriber activity tracking and profiling. While IdP policies and procedures are important in ensuring adherence to appropriate use limitation and purpose specification principles, technical measures such as outlined in 4.1.4 for distributed federation and 5.2.5 for pairwise pseudonymous identifiers, can increase the effectiveness of these policies by making it more difficult to track or profile subscribers beyond operational requirements.
+Section 4.2 also encourages the use of technical measures to provide unlinkability and prevent subscriber activity tracking and profiling. While IdP policies and procedures are important in ensuring adherence to appropriate use limitation and purpose specification principles, technical measures such as outlined in 4.1.4 for distributed federation and 5.2.5 for pairwise pseudonymous identifiers, can increase the effectiveness of these policies by making it more difficult to track or profile subscribers beyond operational requirements.
```
jricher: change "distributed federation" to "proxied federation", and should we add links to sections throughout these paragraphs?

sp800-63c/sec5_assertions.md
```diff
@@ -2,7 +2,7 @@
 
 ## 5. Assertions
 
-An assertion is a packaged set of attribute values and/or attribute claims about an authenticated subscriber that is passed from the identity provider (IdP) to the relying party (RP) in a federated identity system. Assertions contain a variety of information, including assertion metadata, attribute values and claims about the subscriber, and other information that the RP can leverage, such as restrictions, expiration time, etc.
+An assertion is a packaged set of attribute values or attribute claims about or associated with an authenticated subscriber that is passed from the identity provider (IdP) to the relying party (RP) in a federated identity system. Assertions contain a variety of information, including assertion metadata, attribute values and claims about the subscriber, and other information that the RP can leverage, such as restrictions, and expiration time.
```
jricher: "including assertion metadata, attribute values and claims about the subscriber," -> "including assertion metadata, attribute values and attribute claims about the subscriber,"

jricher (changes requested):

sp800-63c/sec10_usability.md
jimfenton: Fixed (in two places), plus a couple random typos

sp800-63c/sec5_assertions.md
jimfenton: Fixed.

sp800-63c/sec5_assertions.md
jimfenton: Would like @garciamike comment since this was his change.

sp800-63c/sec9_privacy.md
jimfenton: Fixed, and section number links added in Section 9.

sp800-63c/sec5_assertions.md
paul-grassi: +1 mostly to @jricher edits. Few thoughts/questions:

* What is sufficient entropy?  Probably need to pull in a minimum here.
* With the introduction of guessing, and the possibility it could use some info to generate keys, do we want to bring in some memorized secret language to help online and offline attacks?
* Is there a compromise to keep @garciamike intent without using 'proxy'?  Perhaps 'use by one IdP-RP pair or those acting in those roles'? Or be generic, like '...use by 1 set of endpoints'?

jricher: @pgrassi-nist great points.
 - I really just mean "random enough" but if we've got a good guideline let's put it in.
 - I think referencing the memorized secrets would help here.
 - The intent was "1 set of endpoints" since the proxy acts as endpoints (as described in the proxy section). I don't want to special case "and also proxies!" throughout the document.

garciamike: I understand the desire to avoid "and also proxies" but I'm not sure the way to get there is by seeming to exclude them in the language we use. I think one set of endpoints might get us there. I'd have to read it in context.

jricher: My argument is that we elsewhere define proxies as acting as an IdP and an RP already. Specifically from 4.1.4:

> Where proxies are used, they function as an IdP on one side and an RP on the other side. Therefore, all normative requirements that apply to IdPs and RPs SHALL apply to proxies in their respective roles.

So we definitely don't need a normative repeat here, and I think even an editorial repeat is redundant.

paul-grassi: Let's get the edit in with 'endpoints', @garciamike can see it in context, and we should be good to go.  I am in agreement to drop excessive proxy language.

jimfenton: What we're trying to do is to prevent parties, such as two RPs, from colluding to associate their interactions with each other based on identifier alone. Random identifiers are one way to do it, but not the only way so I don't see the amount of entropy as much of a factor (as long as you don't have an identifier collision). Another way is to do a keyed hash using the user's "master" (internal) identifier, the RP name (domain?) and some secret key. There are probably other ways  to do this as well.

I'll try to come up with a wording that picks up all of this.

sp800-63c/sec5_assertions.md
```diff
@@ -85,7 +85,7 @@ Note that in a proxied federation model, the initial IdP may be unable to genera
 
 #### 5.2.6. <a name="ppi-gen"></a>Pairwise Pseudonymous Identifier Generation
 
-Pairwise pseudonymous identifiers SHALL be opaque and unguessable, neither containing nor generated from identifying information about the subscriber. Normally, the identifiers SHALL only be known by and used by one IdP-RP pair or IdP-Proxy and Proxy-RP pair. However, an IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, but only if:
+Pairwise pseudonymous identifiers SHALL be opaque, containing no identifying information about the subscriber. They SHALL also be unguessable by a party having access to some information identifying the subscriber. Pairwise pseudonymous identifiers MAY be generated and assigned randomly to subscribers by the IdP or MAY be derived from other subscriber information if the derivation is done in an irreversible, unguessable manner (e.g., using a keyed hash function with a secret key). Normally, the identifiers SHALL only be known by and used by one pair of endpoints (e.g., IdP-RP). However, an IdP MAY generate the same identifier for a subscriber at multiple RPs at the request of those RPs, provided:
```
jimfenton: Hope this covers it. I also didn't think "unguessable" was clear (unguessable to whom?) so I broke that out and defined it; hope that looks OK.

garciamike: This works for me. Thanks, Jim!

jricher: I would change "generated and assigned randomly" to "generated randomly and assigned" but otherwise I'm good with this edit.

garciamike: ha. good catch.

paul-grassi:
@jimfenton we need  `<div class="breaker"/>` before each section for printing purposes.

paul-grassi (approved):

jricher (approved):

PR #473, opened 2017-01-07
Final usability updates for -A

paul-grassi:
@garciamike -A usability is ready for you.  I smoothed some text, got rid of hanging bullets, deleted some OBE KBV statements, and added section references.  Otherwise, retained the body of work from IAD.  Have at it.  @jimfenton I think we have line terminator issues - you are the expert :)  If you don't have the time, I can have at it.


jimfenton:
It looks like we also have issues with quote style (curly quotes vs. straight, with the curly quotes coming from Microsoft Word), which @jricher pointed out some time ago. I'll dig into that as well.

jimfenton (approved):
This looks good. Commits added to fix line terminators, straighten "curly" quotes, and merge mainline nist-pages (due to small conflict in one file).

PR #474, opened 2017-01-07
Ack Ack - Fixed acknowledgments and authors.

paul-grassi:


jimfenton (approved):
Looks good. Note that we'll probably have some conflict resolution to do on the cover sections, because they have changed since this branch was pulled.

PR #488, opened 2017-01-24
Modernize Kerberos documentation

frozencemetery:
This is some language for improvements that have been made to Kerberos, in particular [FAST](http://k5wiki.kerberos.org/wiki/Projects/FAST) and [SPAKE](http://k5wiki.kerberos.org/wiki/Projects/SPAKE_Preauthentication) preauth.

I'm not tied to this specific language or explanation, but I do think they need to be mentioned.  Thanks!

### DIFF
--- a/sp800-63b/sec3_definitions.md
+++ b/sp800-63b/sec3_definitions.md
@@ -158,6 +158,11 @@ initial user-to-KDC exchange. Longer password length and complexity provide
 some mitigation to this vulnerability, although sufficiently long passwords
 tend to be cumbersome for users.
 
+When Kerberos password-based authentication is used in a FAST tunnel, a
+Man-in-the-Middle attack is required in order to perform a dictionary attack
+on the Kerberos protocol. Offline brute force attacks are made impossible
+when SPAKE is used for key exchange.
+
 #### Man-in-the-Middle Attack (MitM)
 An attack on the authentication protocol in which the attacker positions himself or herself in between the claimant and verifier so that he or she can intercept and/or alter data traveling between them.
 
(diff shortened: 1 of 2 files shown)


sp800-63b/sec3_definitions.md
```diff
@@ -147,9 +147,21 @@ A set of attributes that uniquely describe a person within a given context.
 A metric describing degree of confidence that the Applicant’s Claimed Identity is their real identity.
 
 #### Kerberos
```
jimfenton, line 149: Kerberos is not referenced in SP 800-63B, so this definition probably belongs in -63C instead, where the term is used.

The definition is too long already; it should just describe quickly what it is in a sentence or two. The description of how it works belongs in -63C Section 11.2.

sp800-63c/sec11_examples.md
```diff
@@ -55,7 +77,7 @@ subscriber, or in the PKINIT variant of Kerberos, a public key
 certificate. It should be noted that most variants of Kerberos based on
 a shared secret key between the subscriber and IdP derive this key
 from a user generated password. As such, they are vulnerable to offline
-dictionary attack by a passive eavesdropper. 
+dictionary attack by a passive eavesdropper unless SPAKE or FAST is used.
```
jimfenton, line 80: SPAKE and FAST are not included in RFC 4120, so citations to their specifications need to be included. Note that Internet Drafts, as works in progress, are not suitable to be cited here since this document is expected to live much longer than any Internet Draft.

jimfenton (changes requested):
Thanks for your submission; while we are not in a public comment period we wanted to give you feedback on this.

sp800-63c/sec11_examples.md
frozencemetery: Understood.  I will prepare a new version that does not mention SPAKE, but FAST is part of rfc6113.

jimfenton:
If you resubmit this, please note the new guidelines for pull requests: nist-pages is frozen for the public comment period; please base your pull on the post-public-comment branch. Thanks!

PR #492, opened 2017-01-25
SMS updates.

paul-grassi:


sp800-63b/sec5_authenticators.md
```diff
@@ -101,7 +101,7 @@ The purpose of the secret is to securely bind the authentication operation on th
...
+The out-of-band device SHALL be uniquely addressable and communication over the secondary channel SHALL be private. Some voice-over-IP telephone services can deliver text messages and voice calls without the need for possession of a physical device; these SHALL NOT be used for out-of-band authentication. Methods that do not prove possession of a specific device SHALL NOT be used for out-of-band authentication. Mechanisms such as smartphone applications that employ secure communications protocols and uniquely identify the out-of-band device SHOULD be used for out-of-band authentication.
```
jimfenton: The first sentence requires unique addressability and private communication, but at the end of the paragraph this is only a SHOULD. These need to be consistent (and apparently weren't in the original version either, oops).

I really think we need to call out email as being unacceptable here, more strongly than PSTN. I can't think of an example where receipt of email proves possession of a specific device, and we need to be specific about that.

sp800-63b/sec5_authenticators.md
```diff
@@ -119,9 +119,9 @@ If the out-of-band authenticator sends an approval message over the secondary co
...
+If out-of-band verification is to be made using the public switched telephone network (PSTN), the verifier SHALL verify that the pre-registered telephone number being used is associated to a physical device. The verifier then sends the authentication secret to the pre-registered telephone number. Changing the pre-registered telephone number SHALL NOT be possible without two-factor authentication at the time of the change. Verifiers SHALL use known and verifiable routes to deliver the secret. Verifiers SHOULD be aware of indicators such as device swap, SIM change, number porting, or other abnormal behavior *before* using the OOB channel to deliver the authentication secret.
```
jimfenton: The second sentence assumes a particular out-of-band flow, while below (2-3 paragraphs later), we describe three possible flows (added in response to a preview draft comment). At least the second of these is vulnerable to spoofing of the source of the text message from the claimant to the verifier.

jimfenton (changes requested):
Most serious issue is the inconsistency about whether use of a secure channel and unique identification of the device is a SHOULD or a SHALL.

sp800-63b/sec5_authenticators.md
readnotify: Nice work!  Eradicate the ban on SMS by sneaking in changes to get rid of the ban.

Who do you work for?  Are you another one of these phone company or SMS-sending employees Jim ?

readnotify:
These changes are contrary to the SMS Deprecation direction of this standard.

They need to be re-worked to properly express the intent of the SMS Deprecation advice.

sp800-63b/sec5_authenticators.md
paul-grassi: I made it an example in the upcoming commit.  I want to be more outcome driven here and less prescriptive.  So ANY method that does not prove possession of a device, such as email or voip, shall not be used.

jimfenton (approved):

PR #496, opened 2017-01-26
Biometric section re: limiting unsuccesful attempts

enewt:
FIDO is doing something different, but even that is still in flux. That said, I would to propose an aligned approach, including the changes I just proposed to them that would make their approach a requirement for biometrics.

### DIFF
--- a/sp800-63b/sec5_authenticators.md
+++ b/sp800-63b/sec5_authenticators.md
@@ -352,7 +352,13 @@ The biometric system SHOULD implement presentation attack protection (PAD). Test
 
 >Note: Presentation attack detection is being considered as a mandatory requirement in future editions of this guideline.
 
-The biometric system SHALL allow no more than 3 consecutive failed authentication attempts, or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the claimant SHALL be required to use a different authenticator or to activate their authenticator with a different factor such as a memorized secret.
+The biometric system SHALL allow no more than 5 consecutive failed authentication attempts or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the biometric authenticator SHALL enforce either
+
+a) an exponential increase of this delay, e.g. 1 minute after the 6th+ false attempt, 2 minutes after the 7th+ false attempt, 4 minutes after the 8th+, etc. 
+
+**OR**
+
+b) disabling the biometric user verification after the 5th (and subsequent) failed attempt and offering another factor (e.g. a different biometric or a PIN/Passcode if it is not already a required factor) if such alternative method is already implemented. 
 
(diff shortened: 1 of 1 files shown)


sp800-63b/sec5_authenticators.md
```diff
@@ -352,7 +352,13 @@ The biometric system SHOULD implement presentation attack protection (PAD). Test
...
 
-The biometric system SHALL allow no more than 3 consecutive failed authentication attempts, or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the claimant SHALL be required to use a different authenticator or to activate their authenticator with a different factor such as a memorized secret.
+The biometric system SHALL allow no more than 5 consecutive failed authentication attempts or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the biometric authenticator SHALL enforce either
+
+a) an exponential increase of this delay, e.g. 1 minute after the 6th+ false attempt, 2 minutes after the 7th+ false attempt, 4 minutes after the 8th+, etc. 
+
+**OR**
+
+b) disabling the biometric user verification after the 5th (and subsequent) failed attempt and offering another factor (e.g. a different biometric or a PIN/Passcode if it is not already a required factor) if such alternative method is already implemented.
```
jimfenton: Does "different biometric" mean a different sensor, different biometric modality, or perhaps just a different finger?

sp800-63b/sec5_authenticators.md
```diff
@@ -352,7 +352,13 @@ The biometric system SHOULD implement presentation attack protection (PAD). Test
 
 >Note: Presentation attack detection is being considered as a mandatory requirement in future editions of this guideline.
 
-The biometric system SHALL allow no more than 3 consecutive failed authentication attempts, or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the claimant SHALL be required to use a different authenticator or to activate their authenticator with a different factor such as a memorized secret.
+The biometric system SHALL allow no more than 5 consecutive failed authentication attempts or 10 consecutive failed attempts if presentation attack detection meeting the above requirements is implemented. Once that limit has been reached, the biometric authenticator SHALL enforce either
+
+a) an exponential increase of this delay, e.g. 1 minute after the 6th+ false attempt, 2 minutes after the 7th+ false attempt, 4 minutes after the 8th+, etc.
```
jimfenton: What is "this" delay (I.e., what is the initial value)? In your example, you used 1 minute, which seems like a reasonable value. Hopefully the base of the exponential increase is not 1.001 or something like that.

jimfenton (changes requested):
A couple of clarifications needed.

paul-grassi (approved):
Fenton has the right comments, so approving so I can get out of the way of the merge.

sp800-63b/sec5_authenticators.md
enewt: Copy and paste error. It's referring to a 30 second delay after 5 failed attempts.

sp800-63b/sec5_authenticators.md
enewt: I was thinking different modality. Now that you raise it, I recommend that, but it is debateable. I'd be fine with either clarification. Wdyt?

jimfenton (approved):

PR #537, opened 2017-02-07
clarification regarding biometrics

enewt:
Clarification that biometrics are always MFA if follwoing 63 requirements.


paul-grassi:
Closing pull. Need to do PR against post public comment.

PR #541, opened 2017-02-08
Add "Memorized Secret" definition

jweden:
Need definition for "Memorized Secret".  Carried over similar definition of "Memorized Secret Token" from: http://nvlpubs.nist.gov/nistpubs/ir/2015/NIST.IR.8014.pdf

### DIFF
--- a/sp800-63b/sec3_definitions.md
+++ b/sp800-63b/sec3_definitions.md
@@ -159,7 +159,7 @@ When Kerberos authentication is based on passwords, the protocol is known to be 
 An attack on the authentication protocol in which the attacker positions himself or herself in between the claimant and verifier so that he or she can intercept and/or alter data traveling between them.
 
 #### Memorized Secret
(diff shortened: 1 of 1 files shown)


jimfenton:
Thanks for pointing out that we don't have this definition. As for the wording, I lean somewhat toward the definition at the beginning of 5.1.1, in particular that is intended to be memorable by the user. Also, memorized secrets aren't always shared between the subscriber and verifier: a memorized secret that is used to activate a multi-factor crypto authenticator being a counterexample. Is it OK if we substitute the 5.1.1 definition (probably just the first sentence)?

jweden:
Was just looking at 5.1.1 too.  Yeah, that makes sense.

jimfenton:
OK then. Two things:

1. Please also open an issue, so our list of resolved issues is all in one place. We'll use this pull request to close that issue.
2. Your choice, either push another commit to your branch to use the 5.1.1 wording, or I can do that if your branch is set up in the usual way by GitHub. Let me know your preference.

jweden:
Done: https://github.com/usnistgov/800-63-3/issues/542

jimfenton:
Closes #542.

paul-grassi (approved):

PR #563, opened 2017-02-22
Kerberos modernize

frozencemetery:
Hello again, this is the followup to #488 (this time based on the post-public-comment branch).  It only talks about FAST, and I've moved the Kerberos definition as requested in previous review.

If SPAKE is approved by IETF within the year (since an implementation exists now), is there a mechanism by which it could be mentioned in this document, or is that too far out?

Thanks!


jimfenton:
There really isn't a mechanism to make an out-of-cycle revision for something like this. But it's not necessary; the SP 800-63-3 suite is primarily a requirements document: while in some cases it gives examples of how particular protocols (such as Kerberos) may be used, it does not prescribe the use of particular protocols. Even though SPAKE is not mentioned, it may be used as appropriate.

frozencemetery:
Okay, thanks!

sp800-63c/sec3_definitions.md
```diff
@@ -50,6 +50,14 @@ Communication between two systems that relies on redirects through an intermedia
...
 The party that manages the subscriber's primary authentication credentials and issues assertions derived from those credentials. This is commonly the CSP as discussed within this document suite.
 
+#### Kerberos
+
+A widely used authentication protocol developed at MIT. In "classic" Kerberos,
+users share a secret password with a Key Distribution Center (KDC). The user,
+Alice, who wishes to communicate with another user, Bob, authenticates to the
+KDC and is furnished a "ticket" by the KDC to use to authenticate with Bob.
+See section 11.2 for more information.
+
```
jricher: This needs to get moved to the new definitions section.

jricher (changes requested):

jricher (approved):

frozencemetery:
@jricher do you still want changes from me, or did you take care of it?  Thanks!

jricher:
@frozencemetery I took care of it in fa22348 -- it was just putting the changes to the definition in the right file.

Issue #505, opened 2017-01-30
Password hashing/storage recommendation


jethrogb:
**Organization Name (N/A, if individual):** Fortanix, Inc.

**Organization Type (see below for codes):** 2 - Industry

**Document (63-3, 63A, 63B, or 63C):** 63B

**Reference (Include section and paragraph number):** [Section 5.1.1.2](https://pages.nist.gov/800-63-3/sp800-63b.html#memorized-secret-verifiers), paragraph 11

**Comment (Include rationale for comment):**

Iterative salted hashing has long been used in practice for password storage. As far as I'm aware however, there is no NIST-recommended algorithm for password storage. In particular, NIST SP 800-132 “Recommendation for Password-Based Key Derivation” currently mentioned in the referenced section specifically states “This Recommendation specifies a family of password-based key derivation functions (PBKDFs) *for deriving cryptographic keys* from passwords or passphrases *for the protection of electronically-stored data or for the protection of data protection keys*.” (emphasis mine). Clearly, password storage is not within the scope of that recommendation.

It is not immediately obvious that good key-derivation functions are also good password storage mechanisms. (Specifically, I don't think pre-image resistance is a necessary property for a good KDF). Before making such a recommendation, there should be sufficient analysis of the merits.

Additionally, it has become clear that purely hash-based password storage algorithms are easily brute-forced using GPUs or custom hardware. NIST should consider recommending an algorithm that has been specifically designed for password storage in the face of these developments, such as [Argon2](https://password-hashing.net/) or [Balloon](https://crypto.stanford.edu/balloon/).

**Suggested Change:**

Please clarify:

* Is recommending new cryptographic algorithms or new uses of existing algorithms within the scope of this SP?
* Is NIST recommending a cryptographic algorithm for password storage?
* Is NIST recommending that key-derivation functions be used for password storage?
* Is NIST approving PBKDF2 specifically for password storage?
* Will NIST be considering alternative algorithms for password storage that better protect against the threats of today (and the future)?

PR #575, opened 2017-03-01
Improved description of memorized secret storage requirements. Fixes …

jimfenton:
…#505


paul-grassi (approved):

sebastien-rosset:
In 800-63b section "5.1.1.2 Memorized Secret Verifiers", BALLOON and PBKDF2 are provided as "examples" of suitable key derivation functions, without mentioning whether they are memory-hard functions or not. The use of the word "example" in this context means one can expect there would be other suitable functions.

The next sentence states that a "memory-hard function SHOULD be used", without providing a list of suitable key derivation functions that are memory-hard. Isn’t it a contradiction to state that 1) PBKDF2  is suitable and 2) a “memory-hard function SHOULD be used”? Sure, the spec does not state "memory-hard function SHALL be used", so technically one can argue this is not a contradiction, but still, it is perplexing.

The next sentence ("The key derivation function SHALL use an approved one-way function...") gives an explicit list of approved hash algorithms. Was it intentional to provide an explicit list of approved hash algorithms, and an open-ended recommendation for the key derivation function? This leaves the spec open to interpretation.

PR #595, opened 2017-03-13
Fix misspelling in index.md

kas:
Fix misspelling where organizations was spelled "organizatoins"


jimfenton:
This pull request pointed out that the line terminators in this file had been inconsistent: we are using Unix/Linux style line terminators (nl only) while this file had, mostly, cr/nl terminators. My additional commit corrects that.

jimfenton (approved):
Corrects spelling error, also points out that line terminators had been wrong (which is fixed by this pull).

PR #657, opened 2017-03-27
Update of 83C - Section 5.1.2 for clarity

lachellel:
Original contains a mix of descriptive and background information on what is a holder of key assertion (etc) mixed in with the Requirements (shall and may statements).  This can be confusing for the readers to delineate the descriptive from the actual requirements. 

Suggested rearrangement of this sub-section.  No additional requirement specific changes made directly in this PR.  Issues opened for requirement specific changes.


paul-grassi:
@jricher @jimfenton I think these are reasonable updates.  Perhaps the language about unencrypted keys is redundant but it may be worth repeating.  I think a sentence that tees up the bulleted list (and now we need to make bullets numbers) is a good addition.

jricher (approved):
Looks good to me -- doesn't lose any information or requirements and the data flow is better.

PR #1272, opened 2017-03-31
Risk mgmt sync to post pub comm

paul-grassi:
@jimfenton this syncs post-public-comment branch with what we put on nist-pages. No need to review for content since this is out for extended public comment, but make sure I am not screwing anything up.  In my review, all looks good.  All the right files with all the right changes applied to-date on post-pub-comment branch.


jimfenton (approved):
Looks good with AAL to IAL diagram fix.

PR #1660, opened 2017-04-19
Closes #587

### LINKED ISSUE #587: Add "PKI as a Token" to Assertion Examples
### All Fields Are Required
**Organization Name (N/A, if individual)**: DoDMDC

**Organization Type (see below for codes)**: 1
 
**Document (63-3, 63A, 63B, or 63C)**: 63C

**Reference (Include section and paragraph number)**: 11. Assertion Examples
 
**Comment (Include rationale for comment)**:
 
Add a forth assertion technology based on Public Key Infrastructure (PKI) as a token.
 
**Suggested Change**:

Add a forth Section that describes the mechanism for an in-line outbound TLS terminate and regenerate connection capability.  In this framework the outbound perimeter device verifies the users authenticator, up to and including AAL 3, obtains IAL information from the users validated PKI certificate and then requests a short validity period PKI certificate using a private key controlled by the perimeter device to represent the user to the relying party requested by the user.

 
---
 
Organization Type: 1 = Federal, 2 = Industry, 3 = Academia, 4 = Self, 5 = Other

paul-grassi:
@twbaldridge here is some minor text to convey that we are not restricting protocol and payload to things like SAML or OIDC.  

We also plan on building an implementation guide that can get into deployment details.  DoD participating and writing a portion of that to cover certificate assertions may be a good thing.

### DIFF
--- a/sp800-63c/sec11_examples.md
+++ b/sp800-63c/sec11_examples.md
@@ -5,7 +5,7 @@
 
 *This section is informative.*
 
-Three types of assertion technologies will be discussed: SAML assertions, Kerberos tickets, and OpenID Connect tokens. 
+Three types of assertion technologies are discussed below: SAML assertions, Kerberos tickets, and OpenID Connect tokens. This list is not inclusive of all possible assertion technologies, but does represent those that are commonly used in federated identity systems.
 
 ### 11.1. Security Assertion Markup Language (SAML)
 
--- a/sp800-63c/sec5_assertions.md
+++ b/sp800-63c/sec5_assertions.md
@@ -5,7 +5,7 @@
 
 *This section is normative.*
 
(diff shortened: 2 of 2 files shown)


jimfenton (approved):

twbaldridge:
I agree with the update. 

I'm looking forward to the deployment detail development.

The following comment was in #587: Where did this end up?
@jimfenton per our discussion with Tim, let's make sure that FIPS 140 requirements are applied at each FAL (and AAL). Still waiting on response from Andy and Ray, but we know we need to fix that.
I will look at -C to see what the most appropriate language is, if any, to cover this type of assertion generation.

jimfenton:
We just got responses from Andy and Ray today, and I need to review that against what the text currently says.

paul-grassi:
Right.  It's not going to be as easy as a large swath of FIPS application. We are going to need to be conservative so we aren't creating draconian requirements for RPs/IDPs, but we expect some level of HW crypto at the higher FALs.

PR #1667, opened 2017-04-28
Definitions Appendix Update!

paul-grassi:


jimfenton (changes requested):
The renumbering of the sections is very incomplete. If Section 3 is being removed, the section numbers in the following sections need to be changed and the references to section numbers elsewhere in the text need to be updated. It might also be worthwhile to change the names on the source files.

Alternatively, keep section 3 with a reference to the appendix, so that things don't need to be renumbered.

paul-grassi:
That's what I get for multitasking. Crud.  Yes, 3 is staying with a reference.  Standby

paul-grassi:
@jimfenton commit reverted.  This should be good, keeping numbering in tact.

sp800-63c/sec3_definitions.md
```diff
@@ -3,58 +3,4 @@
...
-
-#### Identity Provider (IdP)
-The party that manages the subscriber's primary authentication credentials and issues assertions derived from those credentials. This is commonly the CSP as discussed within this document suite.
-
-#### Pairwise Pseudonymous Identifier
-An opaque unguessable subscriber identifier generated by a CSP for use at a specific individual RP. This identifier is only known to and only used by one CSP-RP pair.
-
-#### Relying Party (RP)
-In this document, the party that receives and processes the assertion identifying the subscriber.
+See [800-63-3, Appendix A](https://pages.nist.gov/800-63-3/sp800-63-3.html#def-and-acr) for a complete set of definitions and abbreviations.
```
jimfenton: It can just say Appendix A because it's the same volume. I will push a fix.

jimfenton (approved):

sp800-63c/sec3_definitions.md
jimfenton: Oops, it's not. My mistake. But I will add the SP.

PR #1772, opened 2017-05-14
Allign with pull #1771

paul-grassi:
replace claim with reference.  @jricher is this nuanced enough where we should denote attribute reference and attribute value in the text? Like `attribute reference` and `attribute value`.

It makes the reader think something is up, rather than us just being casual about attributes.


jricher (approved):

paul-grassi:
@jricher any thoughts on the question in the PR?

jricher:
Unless we do the same with other terms throughout the document, which we currently don't, then I wouldn't call these out specially.

Issue #610, opened 2017-03-15
Federation levels do not represent the strength of the IdP assertion protection


LarryFrank:
### All Fields Are Required
**Organization Name (N/A, if individual):**
N/A
**Organization Type (see below for codes)**:
 Self
**Document (63-3, 63A, 63B, or 63C)**:
63C
**Reference (Include section and paragraph number)**:
 7.1
**Comment (Include rationale for comment)**:
 The discussion seems to assume that all assertions are at common strength "approved cryptography" but nowhere does it discuss the protection of the cryptographic keys of the IdP.  Is it OK to accept an LOA 4 assertion that is signed with what is essentially an "AAL 2" mechanism.
**Suggested Change**:
To meet LOA4, the security of the IdP key used to sign the assertions must meet the same AAL3 level of security as authenticator it sends.   
---
 
Organization Type: 1 = Federal, 2 = Industry, 3 = Academia, 4 = Self, 5 = Other

jricher:
These are two very different uses for keys. Keys used to prove presence of a user need to be unlocked interactively by the user at higher levels. However, what does that mean for the keys used for an IdP? Does there need to be an administrator sitting at the IdP's console to manually unlock its keystore every time a user (who isn't the administrator) needs to log in? 

The IdP's keys obviously do need to be stored and protected appropriately, but how exactly that happens is dependent on the implementation and deployment of the IdP software. We do already implicitly require this in §7.1:

> At any FAL, the IdP SHALL ensure that an RP is unable to impersonate the IdP at another RP by protecting the assertion with a signature and key using approved cryptography.

LarryFrank:
> These are two very different uses for keys.

Agree - the security of the IdP's keys is actually of greater concern than a single user's key.  I read the statement in 7.1 as providing some protection against replay - but not specifying any requirement related to the protection of the IdP's keys.  "Appropriately" is a very ambiguous term when it comes to security of keys.  Appropriate in whose opinion - since it is unlikely the RP actually gets to review the practices of the IdP.

I am not suggesting manual unlocking of the IdP key - but, like a CA, the IdP can have specific security requirements related to the security of the keys (e.g., FIPS 140-2 level 2 hardware).

jimfenton:
Given that we impose FIPS 140 requirements on verifiers (in -63B), it makes sense to have comparable requirements on the protection of private keys by the IdP here.

PR #1791, opened 2017-05-25
Keys and policy updates

jricher:


sp800-63c/sec8_security.md
```diff
@@ -95,7 +95,9 @@ described in the last subsection.
 
 	1.  The assertion is digitally signed by the IdP. The RP 
     checks the digital signature to verify that it was issued by a
-    legitimate IdP.
+    legitimate IdP. The key used is stored in suitably secure storage available to the application 
+	(e.g., keychain storage, trusted platform module, secure element) and is communicated to the RP in 
+	a secure manner (e.g., through a trusted out-of-band mechanism or delivery over a secure protected channel).
```
jimfenton: secure protected channel -> authenticated protected channel
because we have defined the latter term and it includes requirements like use of approved crypto and sufficient key lengths. (also in the other sections you changed below)

jimfenton (changes requested):
I'm not sure this really responsive to #610 -- the changes are in Section 8 which is informative so they're not really binding. The comment makes a good point that we have specific security requirements for FIPS 140 for authenticators and verifiers, and it makes sense to have similarly specific and normative requirements for generators of assertions. But FIPS 140 requirements can be tricky because they sometimes require manual interaction, at least at startup. We should probably discuss this one on a call.

jricher:
The issue was specifically referencing this text in this section (used to be section 7), so that's where I added the discussion.

jimfenton (approved):

PR #1795, opened 2017-05-26
updated privacy discussion for whitelists and TOFU

jricher:


jimfenton (approved):
Looks good to me. Holding off on merge to allow @pgrassi-nist or @nbl-nist to review as well.

paul-grassi (approved):
OK with me.  Give @nbl-nist a chance and will merge by Sunday (maybe Monday).

nbl-nist:
Here are some clarifying edits, and is it clear that “this document” means all of 63 or just 63C?

IdPs MAY establish whitelists of RPs that are authorized to receive authentication and attributes from the IdP, without runtime decision from the subscriber as long as the subscriber is involved in the authentication transaction. IdPs SHALL only place RPs on the whitelist that abide by the provisions and requirements in this document.

jricher:
I'm fine with changing "this document" to something more robust if desired.

jricher:
Also, the subscriber is _always_ involved with the authentication transaction so the additional wording above doesn't add anything explanatory or actionable in my opinion.

nbl-nist:
OK, what is the scope of involved? I mean something akin to the subscriber initiated request for access to resources and the entire chain of SSO is responsive to that initial request. Unless you’re saying that it’s impossible for the IdP to include RPs in the whitelist that don’t have any relationship to the original reason for the subscriber’s log-in, then I’m concerned about what the scope of interpretation for involved is.

jricher:
By "involved" here, I mean that the subscriber is the one sitting at the computer initiating the transaction. The definition of a federation transaction is proving the presence of the subscriber to a party over the network. Not having the subscriber there violates that very basic principle. Either they initiated the call directly -- "log me in to X using Y" -- or indirectly, by taking some action that requires authentication. The latter case is going to be more common with whitelists for the reasons we talked about. But in both cases, the user's there. That's the whole point.

I believe this aligns with your use of "involved" above as well (it was your original word choice, I was copying your usage).

As such, *none* of this applies if the user's not there. You can't send attributes to an RP outside of an authentication transaction (because you're not allowed to), even if the RP is whitelisted. Technologically, of course, we can't stop it. But I think we should remain silent on that point here because it's covered fully by the context of the requirements.

jricher:
@jimfenton You've got language elsewhere that references basically "everything that we say you have to do in 800-63-3", can you dig that up so I can use it here directly? I'd like to be consistent, but I wasn't able to find it on a cursory skim.

jimfenton:
@jricher I'm not sure what you're trying to say. We don't have many back-references from -61A and -63B to -3, so there isn't much to draw from.

jricher:
OK, if it's fine with "this document" then we'll leave it this way, go ahead and merge it.

jricher:
Text updated as per this morning's call.

PR #1800, opened 2017-05-27
A few privacy issues

jimfenton:

Issue #927, opened 2017-03-30
The discussion in section 9.3 is arguably better than the ‘ability to support’ requirement in section 6.3. 


dig-comments:
**Organization Name (N/A, if individual)**: National Security Agency

**Organization Type**: 1
 
**Document (63-3, 63A, 63B, or 63C)**: 800-63C

**Reference (Include section and paragraph number)**: 6.3, 4
 
**Comment (Include rationale for comment)**:  The discussion in section 9.3 is arguably better than the ‘ability to support’ requirement in section 6.3.
 
 Consider amending section 6.3 accordingly to include ‘The RP SHALL minimize the attribute information requested from an IdP to necessary for its access control decision. The IdP SHALL only provide the minimum information requested by the RP. Additional requirements for data retention (specific to aggregation) may also be needed.

**Suggested Change**: 
 
---
 
Organization Type: 1 = Federal, 2 = Industry, 3 = Academia, 4 = Self, 5 = Other

jricher:
RPs need attribute information for much more than access control, though. Display names, email addresses, and other attributes are all important things in an identity federation transaction that don't contribute to access control.

jimfenton:
Will address by adding a forward reference from 6.3 to 9.3.

Also should have a forward reference from the previous paragraph to 9.2 to explain notice.

PR #1801, opened 2017-05-27
Naomi privacy finalize

nbl-nist:


jimfenton:
Don't worry about the links in #927; I committed a fix in #1800 that adds (just) the links.

paul-grassi (approved):

PR #1798, opened 2017-05-26
A copy edit review

RGalluzzo:
Introduction General:
Switching between subject, subscriber, and claimant without clearly
aligning with the graphic. I did not change since it wasn’t clear which
was preferred. Suggest updating to be more consistent as part of the
next revision.
Section 1.2:
For example, in order to maintain integrity of the service, it may be
desirable to know the home ZIP Code of a user for purposes of census
taking or petitioning an elected official but where it is not necessary
or desirable to know the underlying identity of the person. – Awkward
example. Suggest shortening.
Section 2.2:
Real-live v. Real-world. They are used interchangeably. I changed all to
real-life for consistency.
The term validation is not used at all in IAL2 or IAL3. There is also no
discussion of evidence in either of these descriptions. Not sure if this
is intentional.
In the discussion of IAL 2&3 and anonymity there is the following
statement: “It may be beneficial to the user that uses a high assurance
CSP for transactions at lower assurance levels.  For example, a user may
maintain an IAL3 identity, yet should be able to use their CSP for IAL2
and IAL1 transactions.” This does not explain a benefit. I provided some
suggested re-wording.
Section 4:
We state that there is only one outcome of identity proofing here, but
in section one we list 4 outcomes. We should update and harmonize the
language in one of the two sections.
4.1.1.6 Address Confirmation
The term verified and validated seem to be used interchangeably with
regard to the address of record matching what is in records. I am not
sure if this is intentional or if one is preferred.
4.4.1.8 Security controls
Very confusing control. It says they shall have moderate or high and
then says they should have moderate.
Section 5
The table for identity evidence uses a wide array of terms for the
person being proofed – from applicant, to subject, to person, to
individual. I am not sure if this is intentional.

Left off at Table 5-2

PR #1805, opened 2017-05-31
A copy edit review, take 2

jimfenton:
See pull #1798 for conversation and review comments that (mostly) still need to be applied.


RGalluzzo:
RGalluzzo comments from copy edit. These are all on "A."

Introduction General:
Switching between subject, subscriber, and claimant without clearly
aligning with the graphic. I did not change since it wasn’t clear which
was preferred. Suggest updating to be more consistent as part of the
next revision.
Section 1.2:
For example, in order to maintain integrity of the service, it may be
desirable to know the home ZIP Code of a user for purposes of census
taking or petitioning an elected official but where it is not necessary
or desirable to know the underlying identity of the person. – Awkward
example. Suggest shortening.
Section 2.2:
Real-live v. Real-world. They are used interchangeably. I changed all to
real-life for consistency.
The term validation is not used at all in IAL2 or IAL3. There is also no
discussion of evidence in either of these descriptions. Not sure if this
is intentional.
In the discussion of IAL 2&3 and anonymity there is the following
statement: “It may be beneficial to the user that uses a high assurance
CSP for transactions at lower assurance levels. For example, a user may
maintain an IAL3 identity, yet should be able to use their CSP for IAL2
and IAL1 transactions.” This does not explain a benefit. I provided some
suggested re-wording.
Section 4:
We state that there is only one outcome of identity proofing here, but
in section one we list 4 outcomes. We should update and harmonize the
language in one of the two sections.
4.1.1.6 Address Confirmation
The term verified and validated seem to be used interchangeably with
regard to the address of record matching what is in records. I am not
sure if this is intentional or if one is preferred.
4.4.1.8 Security controls
Very confusing control. It says they shall have moderate or high and
then says they should have moderate.
Section 5
The table for identity evidence uses a wide array of terms for the
person being proofed – from applicant, to subject, to person, to
individual. I am not sure if this is intentional.
Table 5-2
“confirming the integrity of physical security features and lack of fraudulent modification.” This is an awkward statement. Suggest re-wording or dropping the “lack of” phrase. A change could be “confirming the integrity of the physical security features and that the document is not fraudulent or inappropriately modified.”
5.3.2
4. Should this read “KBV SHOULD be based on the issuing source or multiple authoritative sources.”?
5. It seems difficult to move to from “should have multiple authoritative sources” to “should use micro-deposits” which don’t really leverage multiple authoritative sources. Perhaps flip 4 & 5 and go with something like “Where transactional history is not used, KBV should be based on multiple authoritative sources.”
6, a. Is this intended to read as if only one (1) of the questions needs to be a free form answer?
5.3.3
Suggest changing “encounter” to “interaction” – encounter sounds unnecessarily aggressive.
5.3.3.2
1.	What qualifies as “departing?” Simply leaving field of view?
2.	“The operator SHALL direct the applicant, as required, to remove any doubt in the proofing process.” Doubt in what? – Suggest rephrasing to “If the remote operator believes the applicant is attempting to fraudulently proof by leveraging means out of their field of view or that would otherwise compromise the validity of the identity proofing session they SHALL ask the applicant to show their surroundings to remove doubt or they SHALL immediately terminate the proofing session.”
5.3.4
3.	Should “subscriber” actually be “trusted referee” here? Otherwise it’s not clear why this would be here.
Section 6
Use case #2. This just seems to be saying that “you want to get a credential.” I am not sure why this is considered “deriving.” It would seem like this should read something like:
An applicant seeks to either add a new authenticator to an existing identity record or establish a new credential at a new CSP leveraging a previously created identity record. For example, if an applicant wants to switch CSPs and/or have another authenticator at a new CSP for other uses (e.g., basic browsing vs. financial). This use case is covered by allowable identity evidence in Section 5.2.
Section 7
“The threats to the enrollment process include impersonation attacks and threats to the transport mechanisms for identity proofing, authenticator binding, and credential issuance. Table 7-1 lists the threats related to enrollment and identity proofing.” This seems to suggest there are more than just the two categories in the previous section. Are there actually three categories of threats: 1) Impersonation, 2) Intercept (mitm? Or something similar), 3) malfeasance of the infrastructure provider?
Denial of enrollment also doesn’t seem to fit neatly into impersonation either. Maybe the statement regarding categories is unnecessary?
Removed repetition of “during issuance” from the “issuance” attacks. Seemed redundant. Also—with these it seems like the examples overlap with MITM and “Duplication” from B. Is the new name for these solely because they are being placed in proofing? Or, do the examples need to be updated. For example:
Disclosure – an individual’s PII being submitted for proofing is collected by an attacker positioning themselves between the CSP and the applicant.
Tampering- An attacker modifies applicant information being submitted to the CSP during registration by positioning themselves between the CSP and applicant.
Mitigation tables:
I am not sure I see much of a difference between the two social engineering examples.
Why combine Unauthorized issuance/social engineering, but not disclosure and tampering?
Section 8
8.1- I would suggest making reference to the benefits of reducing RP risk when discussing data minimization. The benefits as described now are somewhat vague. I am not sure how data minimization equates to more trust in the identity proofing process.
8.1.1-“ Federal agencies need to review any decision to collect the SSN relative to their obligation to reduce the collection and unnecessary use of SSNs under Office of Management and Budget policy.” should probably quote the policy here.
8.6- the use of authentication in this section seems inconsistent with the concept of a “digital identity guideline” rather than “electronic authentication guidelines.”

paul-grassi (approved):

PR #1814, opened 2017-06-01
Garcia final review

garciamike:


garciamike:
Other comments:

Cover: AAL2 definition references approved crypto technique. Do we need to define it here?

Sec1_2: does FAL ONLY refer to assertion protocol or is it more than that?

Sec1_2: table 2-1. Does this undercut the separation into LOA argument too much? Also, we might want to make the table a little easier to interpret, like by adding, "if you have a..." and "...it is comparable to a..." or find some other way to make clear it's a one-way mapping.

Sec1_2: change history. Should we add some past-indicating terms like "then-table 12," etc.? Alternatively, we could pare down the change description to something more generic like the one for -2. 

Sec4: This sentence is odd to me: "therefore, the dashed line between the verifier and the CSP represents a logical link between the two entities rather than a physical link." Is it really a physical link we're talking about?

Sec4: 4.2. last paragraph says "and comply with their responsibilities to the CSP." This is an odd phraes. "Comply with CSP policies in order to maintain active credentials" or something like that?

sp800-63-3/sec5_DIRM.md
```diff
@@ -99,59 +99,69 @@ This section defines the potential impacts for each category of harm. Each assur
...
 
-This guidance addresses only those risks associated with authentication and identity proofing errors. [NIST Special Publication 800-30, "Risk Management Guide for Information Technology Systems"](#SP800-30) recommends a general methodology for managing risk in federal systems. 
+This guidance addresses only those risks associated with authentication and identity proofing errors. [NIST Special Publication 800-30, _Risk Management Guide for Information Technology Systems_](#SP800-30) recommends a general methodology for managing risk in federal systems. 
 
 ### <a name="daps"></a> 5.5. Digital Identity Acceptance Statement
 
-Agencies SHOULD include this information in existing artifacts required to achieve SA&A.
+Agencies SHOULD include this information in existing artifacts required to achieve a security assessment and authorization.
```
paul-grassi: we have the acronym defined already - any reason not to use it here?

paul-grassi (changes requested):

sp800-63-3/sec5_DIRM.md
garciamike: nope. just missed that it wasn't the first use.

paul-grassi (approved):

PR #1820, opened 2017-06-01
Garcia final review 63a

garciamike:


paul-grassi:
Will replace with an updated PR

PR #1832, opened 2017-06-06
More werb

paul-grassi:


jimfenton (approved):
Looks good. Need to explore whether the dx.doi.org URLs in references should also be changed.

PR #1856, opened 2017-06-18
Update sec10_references.md

paul-grassi:
Bad link for DMF


jimfenton (approved):
Approving, but visiting the site indicates that the maintainer is really NTIS, not SSA. Correct if appropriate before merging.

PR #1868, opened 2017-07-19
VoT first draft

jricher:


sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
+<a name="vectors-of-trust-mapping"></a>
 
 <div class="breaker"></div>
 
-## VOT Overview
+## Mapping Assurance Levels to Vectors of Trust
+_This section is normative._
 
-VOT is comprised of components, and components contain values.
+Vectors of Trust (VoT) is a standard for expressing information about an authentication transaction from an IdP to an RP. This information is split into several orthogonal component categories, much in the same way that the assurances defined in 800-63 are split into different xAL categories.
```
paul-grassi, line 8: Is this consistent with C? Should we say transaction or assertion? Or at least fold assertion into this?

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
+### Trustmark
 
-_This section is normative._
+The trustmark URI for these definitions is [[ URI for this document ]].
+
+### xAL to VoT Components
+
+This document uses three categories defined in VoT: Identity Proofing (P), Primary Credential Usage (C), and Assertion Presentation (A). These categories correspond to the Identity Assurance Level (IAL), Authenticator Assurance Level (AAL), and Federation Assurance Level (FAL), respectively. 
+
+This document does not use the Primary Credential Management (M) category defined in VoT.
```
paul-grassi, line 18: I'm thinking we should at M, even if that means an update to B

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
+
+This document does not define any additional component categories. Any additional categories SHALL NOT be used.
```
paul-grassi, line 20: THis is fine, but what happens if iGov defines more? We need some sort of superceding statement, maybe in the doc section that just states 'iGov is approved'.

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
 
 ### IAL to VOT Component Values
 
-May need to break some of these up by remote, remote supervised, in-person.
+IAL maps into the VoT component P using a numeric identifier. Only one numeric identifier SHALL be used. 
 
-IAL maps into the VOT component
+|IAL|Component Value|
+|:----:|:--:|
+|IAL1 (with no attributes)|P0|
```
paul-grassi, line 32: I'm surprised u are ok with this.

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
+|IAL2|P2|
+|IAL3|P3|
 
-|IAL|Options|Component Value|
-|:----:|:---|:--:|
-|IAL1|No attributes|P0|
-|IAL1|Self-asserted attributes|P1|
-|IAL2|Unsupervised Remote and In-Person|P2|
-|IAL3|Supervised Remote or In-person|P3|
+In addition, the type of proofing used MAY be indicated using an alphabetic identifier. Multiple alphabetic identifiers MAY be used simultaneously.
```
paul-grassi, line 37: We will likely need examples, but not necessary now since this is a 'get on the same sheet of music' draft.

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
 
+|Proofing method|Component Value|
```
paul-grassi, line 39: I'm not sure we really need to expose 'sub-sub methods since we deemed them equivalent, but let's consider a vector for:
remote - 1 form of id (there is one case where this is allowed)
in person: supervised remote

And we should have an assertion for 'we've chosen based on risk to be IALx but not implement the requirements'. And maybe a URI to the document that states the compensating controls???

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
-|AAL1||
-|AAL2||
-|AAL3||
+|AAL1|C1|
+|AAL2|C2|
+|AAL3|C3|
+
+In addition, the type of authenticator used SHALL be indicated using an alphabetic identifier. Multiple aphabetic identifiers MAY be used simultaneously.
+
+|Authenticator|Component Value|
```
paul-grassi, line 55: our know our fed friends are going to balk at this. 'we want acronyms!'.

sp800-63d/vot_mapping.md
```diff
@@ -1,54 +1,82 @@
...
+|Single-factor Cryptographic Device|Cd|
+|Multi-factor Cryptographic Software|Ce|
+|Multi-factor Cryptographic Device|Cf|
 
-**Need some language about secondary authenticators since VOT is focused on primary**
 
 ### FAL to VOT Component Values
 
-|IAL|Vector|
+FAL maps into the VoT component A using a numeric identifier. Only one numeric identifier SHALL be used.
```
paul-grassi, line 70: Protocol? As-SAML?Ak-kerb?

paul-grassi (approved):
Love the new grammar check in GH review

sp800-63d/vot_mapping.md
jimfenton: aphabetic -> alphabetic

sp800-63d/vot_mapping.md
jricher: The assertion is passed around in the transaction, but we do mean the transaction here. We'll need to check -C to make sure we get the wording right.

sp800-63d/vot_mapping.md
jricher: There's an implicit universal M throughout all of -3.

sp800-63d/vot_mapping.md
jricher: If iGov defines more then it'll need its own anchor document. That document is allowed to explicitly inherit values from here. We're saying that if you want to do the 800-63-D flavor of VoT then it comes with these categories and values and none others.

sp800-63d/vot_mapping.md
jricher: I had argued initially for an IAL0 equivalent like this, but we folded it into IAL1. This fits with the IETF version of P0 as well, which is a nice parallel.

sp800-63d/vot_mapping.md
jricher: Definitely, examples will need to be everywhere. I didn't want to write examples until we had some idea what we're giving examples of.

sp800-63d/vot_mapping.md
jricher: I'm fine with taking or chopping any of the non-numeric values for each category. They're there as examples and we should be generous with them now and edit them out later.

sp800-63d/vot_mapping.md
jricher: None of these are acronyms, too bad. :-P

sp800-63d/vot_mapping.md
jricher: Not a bad idea, though that's got more to do with the mechanics of presentation than the contents of the assertion. We could try it but my gut says I don't think we'd see much use or value in it. Still, add in now and edit out later.

PR #1874, opened 2017-09-21
private key edits to crypto authenticators.

paul-grassi:


jimfenton:
I pushed a couple of small changes that were in the Word document but not reflected here.

I'm wondering if the following sentence in most of these sections, "The key..." should be plural (key(s) or keys) or should say, "Each such key...". But I haven't made that change because it's not in the Word document either. If you think that's a good idea, please make that change (5.1.6.1 and 5.1.8.1).

jimfenton (approved):

PR #1877, opened 2017-10-13
Proposed component values

kboeckl:
added draft component values for consideration. Have not yet worked on related guidance for these values as the proposed values require further discussion first.
808493d


sp800-63d/vot_mapping.md
```diff
@@ -34,38 +36,70 @@ IAL maps into the VoT component P using a numeric identifier. Only one numeric i
...
 |Remote|Pr|
+|Knowledge-based Verification (KBV)|Pk|
+|Address Confirmation with Postal Address|Pa|
+|Trusted Referee|Pt|
+|Implemented IAL differs from assessed IAL|Px|
 
-### AAL to VOT Component Values
+>Note: Any credential service provider performing identity proofing at IAL2 or IAL3 MUST conduct a privacy risk assessment. Trusted referee policy and procedures MUST be documented per SP 800-63A Section 5.3.4.

+If Px is asserted, the component value for the assessed IAL MUST also be asserted (e.g., “P2.Px”). The discrepancy in assessed and implemented assurance level SHOULD be documented in a digital identity acceptance statement, per SP 800-63-3 Section 5.5.
```
jricher: The numeric level should be a *floor* of the vector component level, any alphabetic bits should satisfy or be additive to the numeric component, not removing capability from it. So instead of P2.Px, you would have P1.Pt.Pk or something like that. An RP can then decide "I'll take any P2 or a P1 only if it has Pt and Pk as well" for a particular action. 

Also, the numeric value is required at all times, whether or not there is an exception. I would recommend that instead of a generic "exception" flag we be more specific, where possible, what the exceptions are and be clear that any exceptions do not remove from the base level definitions.

sp800-63d/vot_mapping.md
```diff
@@ -34,38 +36,70 @@ IAL maps into the VoT component P using a numeric identifier. Only one numeric i
...
+|:----:|:--:|
+|In-person|Pi|
 |Remote|Pr|
+|Knowledge-based Verification (KBV)|Pk|
+|Address Confirmation with Postal Address|Pa|
+|Trusted Referee|Pt|
+|Implemented IAL differs from assessed IAL|Px|
 
-### AAL to VOT Component Values
+>Note: Any credential service provider performing identity proofing at IAL2 or IAL3 MUST conduct a privacy risk assessment. Trusted referee policy and procedures MUST be documented per SP 800-63A Section 5.3.4.
```
jricher: Broken line terminator.

sp800-63d/vot_mapping.md
```diff
@@ -34,38 +36,70 @@ IAL maps into the VoT component P using a numeric identifier. Only one numeric i
...
+|Verifier Impersonation Resistance|Cv|
+|Verifier Compromise Resistance|Cs|
+|Authentication Intent|Cn|
+|Implemented AAL differs from assessed AAL|Cx|
+|Presentation Attack Detection (PAD)|Ck|
+|Biometric comparison performed centrally|Ct|
 
+>Note: Some authentication approaches described in Table XX are not applicable for all three AALs.

If Cx is asserted, the component value for the assessed AAL MUST also be asserted (e.g., “C2.Cx”). The discrepancy in assessed and implemented assurance level SHOULD be documented in a digital identity acceptance statement, per SP 800-63-3 Section 5.5.
```
jricher: Broken line terminator. Same comment about exceptions as above.

sp800-63d/vot_mapping.md
```diff
@@ -75,8 +109,26 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
+
+In addition, the presentation mechanism MAY be indicated using an alphabetic identifier. Multiple aphabetic identifiers MAY be used simultaneously.
 
 |Presentation|Component Value|
+|:----:|:--:|
 |Front channel|Af|
 |Back channel|Ab|
+|Implemented FAL differs from assessed FAL|Ax|
+
+If Ax is asserted, the component value for the assessed FAL MUST also be asserted (e.g., “A2.Ax”). The discrepancy in assessed and implemented assurance level SHOULD be documented in a digital identity acceptance statement, per SP 800-63-3 Section 5.5.
```
jricher: Same comment about exceptions as above.

sp800-63d/vot_mapping.md
```diff
@@ -75,8 +109,26 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
+
+### Credential or Authenticator Management to VoT
+
+800-63 authenticator management maps into the VoT component M using an alphabetic identifier. Authenticator management method MAY be indicated using an alphabetic identifier. Multiple alphabetic identifiers MAY be used simultaneously.
```
jricher: This needs to point to the appropriate chapter.

sp800-63d/vot_mapping.md
```diff
@@ -5,27 +5,29 @@
...
 The trustmark URI for these definitions is [[ URI for this document ]].
 
-### xAL to VoT Components
+### VoT Components
 
-This document uses three categories defined in VoT: Identity Proofing (P), Primary Credential Usage (C), and Assertion Presentation (A). These categories correspond to the Identity Assurance Level (IAL), Authenticator Assurance Level (AAL), and Federation Assurance Level (FAL), respectively. 
+This document uses the four components defined in VoT: Identity Proofing (P), Primary Credential Usage (C), Primary Credential Management (M), and Assertion Presentation (A).
 
-This document does not use the Primary Credential Management (M) category defined in VoT.
+The VoT components P, C, and A correspond to 800-63's Identity Assurance Level (IAL), Authenticator Assurance Level (AAL), and Federation Assurance Level (FAL), respectively. While M does not correspond directly to an 800-63 assurance level category, this document defines a set of M values derived from requirements and recommendations in 800-63.
```
jricher: As stated on call, this section needs to point out that we use the 800-63 terminology in here but reference the VoT terminology where appropriate. So for instance, we'll use "authenticator" here but it will still map to the "C" component, which stands for "Credential" in VoT, but we don't have to call it "Credential" here except for once.

jricher (changes requested):

kboeckl:
@jricher I think I've resolved your comments in these latest commits. And @paul-grassi let me know if there are other changes I should make here.

jricher (approved):

PR #1880, opened 2017-10-22
Offering to fix an errata of sec5_authenticators.md

kthrtty:
Should be fixed?


jimfenton (approved):
This makes it a wee bit clearer. We will merge this when we merge the other errata items; should be very soon now.

PR #1884, opened 2017-11-28
VoT guidance

kboeckl:
Please note, this proposed guidance is for discussion purposes. For example, this version suggests that users should not include values that describe requirements under the asserted xAL, but perhaps we should adjust. I've also bolded other notes throughout for consideration.

### DIFF
--- a/sp800-63d/vot_mapping.md
+++ b/sp800-63d/vot_mapping.md
@@ -5,7 +5,7 @@
 ## Mapping SP 800-63 to Vectors of Trust
 _This section is normative._
 
-Vectors of Trust (VoT) [[VoT]](https://tools.ietf.org/html/draft-richer-vectors-of-trust-05) is a standard for expressing information about an authentication transaction from an IdP to an RP. This information is split into several orthogonal component categories, similar to the way that the assurance levels defined in SP 800-63 are split into different xAL categories.
+Vectors of Trust (VoT) [[VoT]](https://tools.ietf.org/html/draft-richer-vectors-of-trust-07) is a standard for expressing information about an authentication transaction from an IdP to an RP. This information is split into several orthogonal component categories, similar to the way that the assurance levels defined in SP 800-63 are split into different xAL categories.
 
 This document uses terms as defined in [SP 800-63-3 Appendix A](sp800-63-3.html#def-and-acr). Some definitions differ from those in VoT. Notably, the term "authenticator" herein has the same meaning as "credential" in VoT.
 
@@ -17,13 +17,13 @@ The trustmark URI for these definitions is [[ URI for this document ]].
 
 This document uses the four components defined in VoT: Identity Proofing (P), Primary Credential Usage (C), Primary Credential Management (M), and Assertion Presentation (A). This document does not define any additional component categories.
 
-The VoT [component P](https://tools.ietf.org/html/draft-richer-vectors-of-trust-05#section-2.1) corresponds to identity proofing requirements and recommendations found in [SP 800-63A](sp800-63a.html). As such, the document maps SP 800-63's Identity Assurance Level (IAL) and identity proofing characteristics into VoT's P.
+The VoT [component P](https://tools.ietf.org/html/draft-richer-vectors-of-trust-07#section-2.1) corresponds to identity proofing requirements and recommendations found in [SP 800-63A](sp800-63a.html). As such, the document maps SP 800-63's Identity Assurance Level (IAL) and identity proofing characteristics into VoT's P.
 
-The VoT [component C](https://tools.ietf.org/html/draft-richer-vectors-of-trust-05#section-2.2) corresponds to authentication requirements and recommendations found in [SP 800-63B](sp800-63b.html). As such, this document maps SP 800-63's Authenticator Assurance Level (AAL), authenticator types, and authentication characteristics into VoT's C.
+The VoT [component C](https://tools.ietf.org/html/draft-richer-vectors-of-trust-07#section-2.2) corresponds to authentication requirements and recommendations found in [SP 800-63B](sp800-63b.html). As such, this document maps SP 800-63's Authenticator Assurance Level (AAL), authenticator types, and authentication characteristics into VoT's C.
 
-The VoT [component M](https://tools.ietf.org/html/draft-richer-vectors-of-trust-05#section-2.3) corresponds to authenticator lifecycle management requirements and recommendations found in [SP 800-63B Section 6](sp800-63b.html#sec6). As such, this document maps SP 800-63's authenticator lifecycle management characteristics into VoT's M.
+The VoT [component M](https://tools.ietf.org/html/draft-richer-vectors-of-trust-07#section-2.3) corresponds to authenticator lifecycle management requirements and recommendations found in [SP 800-63B Section 6](sp800-63b.html#sec6). As such, this document maps SP 800-63's authenticator lifecycle management characteristics into VoT's M.
 
-The VoT [component A](https://tools.ietf.org/html/draft-richer-vectors-of-trust-05#section-2.4) corresponds to federation requirements and recommendations found in [SP 800-63C](sp800-63c.html). As such, this document maps SP 800-63's Federation Assurance Level (FAL) and assertion presentation into VoT's A.
+The VoT [component A](https://tools.ietf.org/html/draft-richer-vectors-of-trust-07#section-2.4) corresponds to federation requirements and recommendations found in [SP 800-63C](sp800-63c.html). As such, this document maps SP 800-63's Federation Assurance Level (FAL) and assertion presentation into VoT's A.
 
 **Table: 800-63 General Mapping to VoT Components**
 
@@ -36,7 +36,7 @@ The VoT [component A](https://tools.ietf.org/html/draft-richer-vectors-of-trust-
 
 ## Component Values
 
-This document defines new component values in accordance with [VoT Section 2](https://tools.ietf.org/html/draft-richer-vectors-of-trust-05#section-2).
+This document defines new component values in accordance with [VoT Section 2](https://tools.ietf.org/html/draft-richer-vectors-of-trust-07#section-2).
 
 The component values for IAL, AAL, and FAL are expressed with numeric identifiers. Only one numeric identifier SHALL be used for each xAL. The implemented IAL, AAL, and/or FAL SHALL be indicated in the vector. Other digital identity information (i.e., identity proofing characteristics, authenticator type(s) used, authentication characteristics, authenticator lifecycle management characteristics, and assertion presentation values) are expressed with alphabetic identifiers. Multiple alphabetic identifiers MAY be used simultaneously. Some alphabetically-expressed values SHALL be indicated in the vector if implementation matches the uses described in the following sections; otherwise, these values MAY be indicated in the vector.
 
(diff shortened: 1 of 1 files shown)


sp800-63d/vot_mapping.md
```diff
@@ -15,21 +15,19 @@ The trustmark URI for these definitions is [[ URI for this document ]].
...
 
 ## Component Values
 
-This document does not use any of the existing values in VoT but instead defines new component values in accordance with Section 8 of VoT.
+This document defines new component values in accordance with Section 8 of VoT.
 
 ### Identity Proofing
 
-#### IAL to VoT
+IAL maps into the VoT component P using a numeric identifier. Identity proofing characteristics map into P using an alphabetic identifier.
```
jricher: I think these intro paragraphs were better where they were before, because we need to be more specific. Only one numeric, possibly multiple alphabetic, etc. Even though it's all covered below I'd be afraid of that important aspect being missed.

sp800-63d/vot_mapping.md
```diff
@@ -40,26 +38,23 @@ IAL maps into the VoT component P using a numeric identifier. The implemented IA
...
+|**NOTE: We don't think Px conveys very much - because which values represent "approaches beyond those required under the asserted IAL"? Suggest removing this value. Does this comment also apply to Ax for federation?**||
 
 >Note: Any credential service provider performing identity proofing at IAL2 or IAL3 MUST conduct a privacy risk assessment. Trusted referee policy and procedures MUST be documented per SP 800-63A Section 5.3.4.
 
-Any discrepancies in the assessed and implemented assurance level SHOULD be documented in a digital identity acceptance statement, per SP 800-63-3 Section 5.5.
-
 ### Authenticator Usage
 
-#### AAL to VoT
+AAL maps into the VoT component C using a numeric identifier. Authenticator types and characteristics map into C using an alphabetic identifier.
```
jricher: Same comment as above.

sp800-63d/vot_mapping.md
```diff
@@ -40,26 +38,23 @@ IAL maps into the VoT component P using a numeric identifier. The implemented IA
...
-|Identity Proofing Method or Approach|Component Value|
+|Identity Proofing Characteristic|Component Value|
 |:----:|:--:|
 |In-person|Pi|
 |Remote|Pr|
 |Knowledge-based Verification (KBV)|Pk|
 |Address Confirmation with Postal Address|Pa|
 |Trusted Referee|Pt|
 |Additional features implemented beyond requirements for asserted IAL|Px|
+|**NOTE: We don't think Px conveys very much - because which values represent "approaches beyond those required under the asserted IAL"? Suggest removing this value. Does this comment also apply to Ax for federation?**||
```
jricher: Worth keeping in draft form to see if we get feedback on it. I also don't think it adds much on its own but I can possibly see a desire for a general "I'm better than that" flag -- at least that was the original argument.

sp800-63d/vot_mapping.md
```diff
@@ -101,14 +92,23 @@ The authentication characteristics MAY be indicated using an alphabetic identifi
...
+|Authenticator issued/bound after proofing session (remote)|Mr|
+|Authenticator issued/bound after proofing session (in-person)|Mi|
+|Second factor bound to a single-factor account|Ms|
+|Authentication factors reestablished (account recovery) using abbreviated proofing process|Ma|
 
-#### FAL to VoT
+### Federation and Assertions
 
-FAL maps into the VoT component A using a numeric identifier. Only one numeric identifier SHALL be used.
+FAL maps into the VoT component A using a numeric identifier. Assertion presentation maps into A using an alphabetic identifier.
```
jricher: Same comment as above.

sp800-63d/vot_mapping.md
```diff
@@ -118,25 +118,87 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
 |:----:|:--:|
 |Front channel|Af|
 |Back channel|Ab|
 |Additional features implemented beyond requirements for asserted FAL|Ax|
 
-Any discrepancy in the assessed and implemented assurance level SHOULD be documented in a digital identity acceptance statement, per SP 800-63-3 Section 5.5.
+## How to Use the Values in Vectors
 
-### Authenticator Management to VoT
+The implemented IAL, AAL, and/or FAL component value(s) SHALL be indicated in the vector. Only one numeric identifier SHALL be used for each xAL. Any discrepancies in the assessed and implemented assurance level SHOULD be documented in a digital identity acceptance statement, per SP 800-63-3 Section 5.5.
```
jricher: While this covers the combination requirements, I'm a little wary of it being only found down here. Maybe more examples in each section would help?

sp800-63d/vot_mapping.md
```diff
@@ -118,25 +118,87 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
-|Credential Management Method|Component Value|
-|:----:|:--:|
-|Credential issued/bound during proofing session|Mp|
-|Credential issued/bound after proofing session (remote)|Mr|
-|Credential issued/bound after proofing session (in-person)|Mi|
-|Second factor bound to a single-factor account|Ms|
-|Authentication factors reestablished (account recovery) using full proofing process|Ma|
-|Authentication factors reestablished (account recovery) using abbreviated proofing process|Mb|
\ No newline at end of file
+Identity proofing characteristics, authenticator type(s) used, authentication characteristics, and assertion presentation values SHOULD be indicated only when they demonstrate the implementation of approaches beyond those required by the xAL indicated in the vector. The appropriate value that signifies this (i.e., Px, Cx, and Ax) SHOULD also be indicated. Values SHOULD NOT be indicated that represent approaches already required by the xAL indicated in the vector. These approaches are assumed to be implemented per the xAL's requirements.
```
jricher: ok, if this is the only use case for the `x` values then yes it's useless -- I had assumed it was "there's something more that I don't have a more specific thing to talk about"

sp800-63d/vot_mapping.md
```diff
@@ -118,25 +118,87 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
\ No newline at end of file
+Identity proofing characteristics, authenticator type(s) used, authentication characteristics, and assertion presentation values SHOULD be indicated only when they demonstrate the implementation of approaches beyond those required by the xAL indicated in the vector. The appropriate value that signifies this (i.e., Px, Cx, and Ax) SHOULD also be indicated. Values SHOULD NOT be indicated that represent approaches already required by the xAL indicated in the vector. These approaches are assumed to be implemented per the xAL's requirements. 
+
+**Example:** C2.Cx.Cv could be asserted if implementing AAL2 and an authenticator is used that provides verifier impersonation resistance, since verifier impersonation resistance is not required under AAL2.
+
+Any additional component categories SHALL NOT be used.
+
+### Identity Proofing Characteristics
+
+If implementing IAL2 and KBV is used for identity verification, Pk SHALL be indicated in the vector (e.g., P2.Pk).
```
jricher: Is there reason for these normative requirements to be in another section? Seems like this should be bundled with the definitions above, but maybe I'm missing something in the structure.

sp800-63d/vot_mapping.md
```diff
@@ -118,25 +118,87 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
+If using a biometric for authentication, and:
+
+* biometric comparison is performed centrally (not locally on the claimant's device), Ck SHALL be indicated in the vector.
+* the biometric system implements PAD, Ck SHOULD be indicated in the vector.
+
+**NOTE: Should we add a vector value for FMR?**
+
+MitM attack resistance
+
+**NOTE: I'm struggling to find the case where a MitM resistance value is interesting - it seems to be required most of the time and would be redundant with the xAL indicated in the vector.**
```
jricher: I think it's interesting when used with C1

sp800-63d/vot_mapping.md
```diff
@@ -118,25 +118,87 @@ FAL maps into the VoT component A using a numeric identifier. Only one numeric i
...
+
+If a subscriber loses all authentictors of a factor necessary to complete multi-factor authentication, has been identity proofed at IAL2 or IAL3, and consequently undergoes an abbreviated identity proofing process to reestablish authentication factors, Ma SHALL be indicated in the vector.
+
+### Assertion Presentation
+
+If a front-channel presentation mechanism is used, Af SHOULD be indicated in the vector.
+
+If a back-channel presentation mechanism is used, Ab SHOULD be indicated in the vector.
+
+**NOTE: We could build a table to help readers distinguish between values that are required in the vector (SHALLs/depending on implementation) vs. encouraged (SHOULDs/depending on implementation). I can also add in references to A-C as relevant/appropriate so readers can easily find related guidance.**
```
jricher: References are definitely good and a table might help wade through the sea of requirements. In other spaces I've seen formatted lists instead, but that's also using the IETF normative syntax which gives use "REQUIRED" and "OPTIONAL" in addition to "SHALL" and "SHOULD"

jricher (changes requested):

jricher:
Overall the content is good, I think it would benefit from some reorganization. Specifically, fitting it in to the overall structure more.

paul-grassi:
Please address Justin's comments and we can re-review. Nice job.

jricher:
An additional comment: The management section (the `M` vector component) needs to be more clearly stated to be talking about an ongoing *policy* with the account and not signaling an *event* on the account.

sp800-63d/vot_mapping.md
kboeckl: @jricher When you say 'then yes it's useless', are you referring to all of the x values, or just Px (since you suggested leaving Px in the document for feedback in an earlier comment)? Either way, I'd be happy to leave all of the x values in and see what kind of feedback we get. Or we could just remove Px.

sp800-63d/vot_mapping.md
kboeckl: @jricher My understanding is that C1 would imply MitM resistance since it's required at AAL1. Or what is the interesting aspect to C1 with this value as well?

sp800-63d/vot_mapping.md
jricher: Ok, you're right -- I was thinking that it was a SHOULD at AAL1, but if it's required across all AAL's then we should remove this value.

sp800-63d/vot_mapping.md
jricher: Ok, you're right -- I was thinking that it was a SHOULD at AAL1, but if it's required across all AAL's then we should remove this value.

sp800-63d/vot_mapping.md
jricher: My understanding was that `Px`, `Cx`, and `Ax` were to signal that there was something else you did that wasn't covered by another element. Basically, "we did IAL1 and also some other stuff". If you're only allowed to use it when you can also describe exactly what you did, as the current text suggests, then it's redundant and doesn't really add to the expressiveness. I'd say drop all of the `x` categories in that case, which has the nice side effect of making it simpler.

kboeckl: Got it, thanks, that makes sense. @paul-grassi This is probably worth a larger discussion. Perhaps the x values are more interesting if they don't have to be asserted along with the value that demonstrates exactly what they did - but in that case, I would think they should have to explain somewhere why they think they're going beyond their asserted level. And that step would certainly add an additional level of complication.

kboeckl:
@jricher & @paul-grassi: I made a bunch of changes; please review when you have time. Thanks!

jricher (approved):

PR #1891, opened 2018-08-14
make pages slightly more responsive

smiller171:
Some pages, specifically https://pages.nist.gov/800-63-3/sp800-63b.html, render really poorly on mobile due to x-axis overflows. This PR makes some minor changes to improve the situation, though the NavBar should probably be hidden or moved off the left edge to improve things further.

### DIFF
--- a/static/css/NISTStyle.css
+++ b/static/css/NISTStyle.css
@@ -84,7 +84,8 @@ sub{
 }
 
 img{
-	border:0
+	border:0;
+	max-width: 100%;
 }
 
 svg:not(:root){
@@ -318,7 +319,9 @@ input,button,select,textarea{
 
 a{
 	color:#0000FF;
-	text-decoration:none
+	text-decoration:none;
+	overflow-wrap: break-word;
+  	word-wrap: break-word;
 }
 
 a:hover,a:focus{
@@ -1634,7 +1637,9 @@ pre code{
 
 table{
 	max-width:100%;
-	background-color:transparent
+	background-color:transparent;
+	display: block;
+	overflow: scroll;
 }
 
 th{


static/css/NISTStyle.css
```diff
@@ -1634,7 +1637,9 @@ pre code{
 
 table{
 	max-width:100%;
-	background-color:transparent
+	background-color:transparent;
+	display: block;
```
jricher, line 1641: Changing the display of a fundamental element like `table` seems like a hack to to me, what's the driving reason behind this?

jricher (changes requested):

static/css/NISTStyle.css
smiller171: Table as an element isn't designed for responsiveness. This allows the table's width to be confined to the width of the available horizontal space, and allows the table itself to be scrolled instead of scrolling the whole page, which looks much better from a mobile device.

This is the easiest way I personally know how to accomplish this purely in CSS, and I haven't looked at the markup in this repo to know how easy it would be to go through wrapping these tables with div elements or assigning classes to them.

jricher: Thanks for the reply. Changing the markup of tables throughout the document isn't really an option at this stage, so you're right to do something in CSS (or JS as an option) to make it work across the whole document repository. I'm still wary of it but I think we can try it and see if it breaks anything else (like large format displays or print rendering), or if it handles any of our tables oddly. I do remember in making the document we had trouble getting the tables to look exactly like they're supposed to in a NIST document, which is part of why I'm skeptical of a change like this.

jricher (approved):

jimfenton:
Change reverted by #1892 -- see comments there for explanation.

PR #1936, opened 2019-06-14
added docker support

jricher:


jimfenton (approved):
Nice capability!

jricher:
I recommend we cherrypick this into the `nist-pages` branch as well.

jimfenton:
With the exception of the template change, I expect that the next change to the nist-pages branch will be the merge of the errata-2 branch. So I'm not sure it's worth cherry picking.

PR #1939, opened 2019-06-21
Remove "comment period closed" issue template.

brodygov:
It should already be obvious to submitters that the comment period has
long since closed, and this discourages people from submitting new
issues.

Edits are welcome on this PR.


jricher:
Thank you, we will replace this with more appropriate text instead of simply removing it.

Issue #54, opened 2016-05-31
Map FAL to concrete protocols


jricher:
Perhaps we should map the FAL to specific deployment choices of the example protocols in the appendix, to give developers a (non-normative?) hold on what we're expecting.

PR #1941, opened 2019-06-21
add anchor for risk assessment section for link from FAQ

jricher:


jimfenton (approved):
It turns out that Jekyll creates anchors for all the section names as well -- this one is #54-risk-acceptance-and-compensating-controls

But this is shorter and easier to use.

PR #1949, opened 2019-08-16
mark all sub-pages to not be published separately 

jricher:
Flag all relevant markdown pages in subsections of the site to not be rendered to separate HTML pages by Jekyll, since they are included in the index files by hand. This action cleans up the output of the 800-63-3 site to include only publicly accessible pages as intended, instead of having dangling unlinked pages like the following showing up unintended:

https://pages.nist.gov/800-63-3/sp800-63c/sec11_examples.html

### DIFF
--- a/sp800-63-3/cover.md
+++ b/sp800-63-3/cover.md
@@ -1,3 +1,6 @@
+---
+published: false
+---
 <div class="text-right" markdown="1">
 
 # <a name="800-63-3"></a> NIST Special Publication 800-63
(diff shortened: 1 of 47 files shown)


jimfenton (approved):
Looks good; doesn't make sense to publish each chapter as a separate file that isn't referenced anywhere.

jricher:
It turns out that this has a side effect of putting frontmatter into the included file, so this patch needs to be backed out and a new solution found.

PR #1950, opened 2019-08-27
Unpublish included pages

jricher:


jimfenton (changes requested):
As far as I can tell, this is still generating individual HTML pages by major section. Can you check that out? If the changes in your second commit aren't having the desired effect, I'd rather leave them out.

jricher:
I missed these comments -- the pages being generated are only the ones we intend it to. I'm not sure what other individual pages you're seeing, but this commit is doing the right thing and should be merged.

jimfenton (approved):
I'm not sure what I was seeing before, but it seems right now.

PR #1952, opened 2019-10-17
switch to dynamic header and footer for NIST Pages

jricher:


jimfenton (changes requested):
Testing with Docker, I'm not seeing the NIST headers and footers at all.

jricher:
Works fine in my testing environment, and the changes are definitely in place.

jimfenton (approved):
Problem testing in Docker turned out to be that I was using http://0.0.0.0:4000/800-63-3/ rather than http://localhost:4000/800-63-3/

PR #1993, opened 2021-09-06
Removed double space

RufusJWB:
There were two surplus "blank spaces"


jricher:
Thank you for your pull request, but we are not taking changes to revision 3 at this time.

PR #2013, opened 2025-08-26
Change header to refer readers to current guidelines

jimfenton:


jricher:
This looks fine, but do we want to add that to the individual volume pages as well?

jricher:
Adding this line to `_layouts/page.html` would probably do the trick:

```

<div class="alert alert-warning">

This revision of NIST SP 800-63 has been superseded by <a href="https://pages.nist.gov/800-63-4/">NIST SP 800-63-4</a> as of August 1, 2025. Please refer to those documents for the current guidelines.

</div>
```

jimfenton:
Might be a good idea to add those headers to the individual pages, but I don't think I have a way to test the nist-pages rendering locally and I'm less confident of getting that right on the first try. Do you know a way of testing that?

Ryang03:
I think it would be good to include a banner or "superseded" statement in the volumes themselves. But, lets start by getting this one up.

Ryang03 (approved):

jricher:
banner added to all pages in #2014

